PR #121, opened 2022-11-03
Fix interactive shell argument

jonchang:
Only zsh permits the long `--interactive` argument. Tested against:
* bash
* dash
* csh
* zsh

Fixes https://github.com/teaxyz/cli/issues/119.

### DIFF
--- a/src/app.exec.ts
+++ b/src/app.exec.ts
@@ -269,7 +269,7 @@ async function repl(installations: Installation[], env: Record<string, string>) 
   console.info(pkgs_str())
   console.info("when done type: `exit'")
   const shell = Deno.env.get("SHELL")?.trim() || "/bin/sh"
-  const cmd = [shell, '--interactive']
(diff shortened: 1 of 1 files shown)


jhheider (approved):
likely the correct change. I wonder if we can easily write either some unit tests or CI tests to ensure this. Might be hard to test the REPL, though. That sounds like an `expect` thing.

jonchang:
In terms of testing, I think expect could work well. We could also spawn a thread that sleeps for e.g. 5 seconds, then try to launch the interactive process (and check for launch errors: `/bin/bash --interactive` returns a positive exit code), then kill the process upon timeout.

jonchang:
I'm also looking through the shell invocation stuff and there are a number of places in the code that could be improved, so I might work on that later.

mxcl:
Brilliant. Thanks Jonathan.

PR #125, opened 2022-11-04
Update README.md

zohairhadi:
just a smol grammar update


jhheider:
I consider "we're new software" more idiomatic than "we're a new software", but they mean the same thing. I defer to @mxcl 's preference on this one.

mxcl:
Yeah. Sorry. I don’t agree with this change.

PR #127, opened 2022-11-05
Tea served by Gitpod

abitrolly:
Get `./tea` with a single click.

**Click Me:** https://gitpod.io/#https://github.com/teaxyz/cli/pull/127

This downloads `deno` for compiling `tea` into `.deno/` in current dir, because Gitpod doesn't persist $HOME when initializing.
```
Deno was installed successfully to .deno/bin/deno
```
Then it compiles `./tea` using README instructons, but doesn't install it.
```
Download ⠋ https://deno.land/x/rimbu@0.12.3/sorted/map-custom/implementation/immutable.tsCheck file:///workspace/teaxyz-cli/src/app.ts
Download ⠸ https://deno.land/x/rimbu@0.12.3/sorted/map-custom/implementation/immutable.ts                                 Compile file:///workspace/teaxyz-cli/src/app.ts
Emit ./tea
./tea is hot
```

Running `./tea` without arguments fails for some reason.
```
$ ./tea
error: Uncaught (in promise) "not-found:deno.land"
```
But help works fine.
```
$ ./tea --help
usage:
  tea [-xdX] [flags] [+package~x.y] [file|URL|target|cmd|interpreter] -- [arg…]
...
```

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,30 @@
+# List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
+
+tasks:
+  - init: |
+      # runs during prebuild
+      (
+        set -e
+
+        # install deno
+        curl -fsSL https://deno.land/install.sh | sh
+      )
+    command: |
+      # runs during startup
+      (
+        set -e
+
+        # compile ./tea
+        deno compile \
+          --allow-read \
+          --allow-write \
+          --allow-net \
+          --allow-run \
+          --allow-env \
+          --unstable \
+          --import-map="import-map.json" \
+          --output "./tea" \
+          "src/app.ts"
+
+        echo "./tea is hot"
+      )
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -6,8 +6,9 @@ tasks:
       (
         set -e
 
-        # install deno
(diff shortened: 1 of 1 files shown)


mxcl:
> Running ./tea without arguments fails for some reason.
> 
> $ ./tea
> error: Uncaught (in promise) "not-found:deno.land"

without the installer you need to do `tea --sync` first. I should fix the error.

Any reason we don’t use our installer? It doesn't do much, but means we use the precompiled binaries for what we support. Compiling from source may support more platforms, but in practice we support all the platforms Deno supports already.

Also can this file not be in the root? I like what Gitpod is doing and enjoy supporting it, but I don't like files in the root since it clutters the checkout confusing new potential contributors. `.github` has become a defecto place for these even though it makes no sense. Honestly `.git/pod.yml` would be safe though perhaps would be shunned.

abitrolly:
> Any reason we don’t use our installer?

No other reason than to bootstrap `tea` from raw source without anything else except the compiler.

> Also can this file not be in the root?

I don't think so. At least the reference is silent about auto-detection - https://www.gitpod.io/docs/references/gitpod-yml

Unless there is some standard on `.project/` files, I doubt there will be changes. Filled a feature request here https://github.com/gitpod-io/gitpod/issues/14476 Let's see how it goes.

PR #129, opened 2022-11-06
ci: add test for executable markdown

abitrolly:


.github/workflows/ci.yml
```diff
@@ -28,6 +28,19 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: teaxyz/setup@v0
+      - run: tea .
```
mxcl, line 42: Needs to be `tea run .` or it won’t run with the current source code. Our `run` target runs `deno` with the working directory sources.

mxcl (changes requested):

.github/workflows/ci.yml
mxcl: Also you may need to rebase since I just fixed the “cannot write” error.

abitrolly: But documentation says `tea .` should work  - https://github.com/teaxyz/cli#executable-markdown
```sh
$ tea .  # interprets `# Getting Started`, could also be `tea ./README.md`
```

abitrolly: Rebased.

mxcl: `run` is an exe/md target in our `README`.

mxcl (approved):

Issue #130, opened 2022-11-06
cannot run teaxyz/www example an other tea . issues


davidtai:
In https://github.com/teaxyz/www, when I run:
```
sh <(curl tea.xyz)
tea +git-scm.org git clone https://github.com/teaxyz/www tea-www
cd tea-www
tea .
```

I get an issue with temporary files:

```
dtai@Striker-Noir:~/projects/tea-www$ tea -vvv .
{
  args: { args: [ "." ], pkgs: [] },
  flags: {
    verbosity: 3,
    magic: true,
    json: false,
    numpty: false,
    sync: false,
    verbose: true,
    debug: true,
    muggle: false,
    silent: false
  }
}
/home/dtai/.tea/tea.xyz/v0.11.6/bin/tea
error: Uncaught (in promise) "file-exists:/home/dtai/.tea/tea.xyz/tmp/tea1ebdbfb5"
```

This occurs in WSL Ubuntu and OSX.

PR #131, opened 2022-11-06
Basic Path Sanitization

davidtai:
Do some basic path sanitization to prevent tmp escape attempts and to make `tea .` work properly by not trying to overwrite the temporary folder and fixing #130.


mxcl:
Oh interesting. Well I think the proper fix is in `Path.ts` `join`. Which is mine and thus my bad code.

mxcl:
I did wonder why `tea .` would error in some cases!

davidtai:
> Oh interesting. Well I think the proper fix is in `Path.ts` `join`. Which is mine and thus my bad code.

There still needs to be a default filename for `tea .` sh that's set in here(called 'sh' in my fix).  I would normally just pull in a npm sanitization library to prevent any sort of escape but I'm not very familiar with deno's standard libraries.

davidtai:
We should also pull all examples into the CI system to catch stuff like this.  I believe that @abitrolly started doing that in #129.

mxcl:
Per our README `.` means “execute the default target in README.md”.

Perhaps this is dumb, what is your reaction?

davidtai:
That is fine.  My point was directed at there still needs to be a filename appended to the tmp directory as most software will join a '.' as a no op and try and overwrite the directory, 

Since you are here I'd also like to point out that there is a related issue in #133 that is also a somewhat concerning user space execution security issue.  `tea .` is fine and convenient but we just have to make sure it handles all cases, if the markdown is empty:

```
# Getting Started

# sh

## hello
```

Then we try and execute `.` directly because there's an unreachable error message that seems like it should tell the user that there's no commands in `# Getting Started`.  The security aspect of this is what happens when you replace the harmless `.` with something less benign like execute `tea sh` which will open a shell (tested & confirmed).

PR #147, opened 2022-11-10
Store etag So Downloads Use Cache

kunalgoyal9:
Storing Etag received from `fetch` API parallel to mtime file and passing this header to `If-None-Match`. 
Fixing https://github.com/teaxyz/cli/issues/85

### DIFF
--- a/README.md
+++ b/README.md
@@ -802,7 +802,7 @@ deno test \
  --allow-env=SRCROOT,GITHUB_TOKEN,TMPDIR,TEA_PREFIX,MUGGLE,MAGIC,JSON,NUMPTY \
  --allow-run \
  --import-map=$SRCROOT/import-map.json \
- --allow-write \
+ --allow-write=$TEA_PREFIX/tea.xyz/var,$TMPDIR \
  --unstable \
  "$SRCROOT"/tests/**/*.test.ts
 ```
--- a/README.md
+++ b/README.md
@@ -786,10 +786,10 @@ export TMPDIR=${TMPDIR:-/tmp}
 deno test \
  --allow-net \
  --allow-read \
- --allow-env=SRCROOT,GITHUB_TOKEN,TMPDIR,TEA_PREFIX \
+ --allow-env=SRCROOT,GITHUB_TOKEN,TMPDIR,TEA_PREFIX,MUGGLE,MAGIC,JSON,NUMPTY \
  --allow-run \
  --import-map=$SRCROOT/import-map.json \
- --allow-write="$TMPDIR" \
+ --allow-write \
  --unstable \
  "$SRCROOT"/tests/**/*.test.ts
 ```
--- a/src/hooks/useDownload.ts
+++ b/src/hooks/useDownload.ts
@@ -45,10 +45,18 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
   dst ??= hash().join(src.path().basename())
   if (src.protocol === "file:") throw new Error()
 
-  if (!ephemeral && mtime_entry().isFile() && dst.isReadableFile() && etag_entry().isFile()) {
+  if (!ephemeral && dst.isReadableFile()) {
+    
     headers ??= {}
-    headers["If-Modified-Since"] = await mtime_entry().read()
-    headers["If-None-Match"] = await etag_entry().read()
+    if(mtime_entry().isFile() && etag_entry().isFile()){
+      headers["If-Modified-Since"] = await mtime_entry().read()
+      headers["If-None-Match"] = await etag_entry().read()
+    } else if(mtime_entry().isFile()){
+      headers["If-Modified-Since"] = await mtime_entry().read()
+    } else if(etag_entry().isFile()){
+      headers["If-None-Match"] = await etag_entry().read()
+    }
+
     logger.replace(teal('querying'))
   } else {
     logger.replace(teal('downloading'))
@@ -79,7 +87,6 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
       dst.parent().mkpath()
       await body(reader, f, sz)
 
-      //TODO etags too
       const text = rsp.headers.get("Last-Modified")
       const etag = rsp.headers.get("etag")
 
--- a/src/hooks/useDownload.ts
+++ b/src/hooks/useDownload.ts
@@ -48,9 +48,9 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
   if (!ephemeral && dst.isReadableFile()) {
     
     headers ??= {}
-    if(etag_entry().isFile()){
+    if (etag_entry().isFile()) {
       headers["If-None-Match"] = await etag_entry().read()
-    } else if(mtime_entry().isFile()){
+    } else if (mtime_entry().isFile()) {
       headers["If-Modified-Since"] = await mtime_entry().read()
     }
 
--- a/src/hooks/useDownload.ts
+++ b/src/hooks/useDownload.ts
@@ -39,14 +39,16 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
     return () => memo ?? (memo = hash_key(src))
   })()
   const mtime_entry = () => hash().join("mtime")
+  const etag_entry = () => hash().join("etag")
 
   const { numpty } = useFlags()
   dst ??= hash().join(src.path().basename())
   if (src.protocol === "file:") throw new Error()
 
-  if (!ephemeral && mtime_entry().isFile() && dst.isReadableFile()) {
+  if (!ephemeral && mtime_entry().isFile() && dst.isReadableFile() && etag_entry().isFile()) {
     headers ??= {}
     headers["If-Modified-Since"] = await mtime_entry().read()
+    headers["If-None-Match"] = await etag_entry().read()
     logger.replace(teal('querying'))
   } else {
     logger.replace(teal('downloading'))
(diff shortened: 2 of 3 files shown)


src/hooks/useDownload.ts
```diff
@@ -39,14 +39,16 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
...
   })()
   const mtime_entry = () => hash().join("mtime")
+  const etag_entry = () => hash().join("etag")
 
   const { numpty } = useFlags()
   dst ??= hash().join(src.path().basename())
   if (src.protocol === "file:") throw new Error()
 
-  if (!ephemeral && mtime_entry().isFile() && dst.isReadableFile()) {
+  if (!ephemeral && mtime_entry().isFile() && dst.isReadableFile() && etag_entry().isFile()) {
```
mxcl, line 48: This means we'll only do these if BOTH etag and mtime are provided. ETAGs are much rarer than mtimes.

kunalgoyal9: Can we do this if any of mtime or etag condition met?

mxcl: yes, if any of them are met. I recommend reading up on how HTTP caching works, I do not have time to teach you. Thanks.

mxcl:
Needs a test

src/hooks/useDownload.ts
```diff
@@ -39,14 +39,24 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
...
   const { numpty } = useFlags()
   dst ??= hash().join(src.path().basename())
   if (src.protocol === "file:") throw new Error()
 
-  if (!ephemeral && mtime_entry().isFile() && dst.isReadableFile()) {
+  if (!ephemeral && dst.isReadableFile()) {
+    
     headers ??= {}
-    headers["If-Modified-Since"] = await mtime_entry().read()
+    if(mtime_entry().isFile() && etag_entry().isFile()){
```
mxcl, line 51: https://stackoverflow.com/a/10671877

Don’t do both. Please research the task at hand before attempting it.

mxcl (changes requested):
Not following best practices.

tests/unit/cache.test.ts
```diff
@@ -0,0 +1,24 @@
+import { assert } from "deno/testing/asserts.ts"
+import { useDownload } from "hooks"
+import { sandbox } from "../utils.ts"
+
+Deno.test("etag-mtime-check",async () => {
+  await sandbox(async tmpdir => {
+    const src = new URL("https://dist.tea.xyz/deno.land/darwin/aarch64/v1.27.1.tar.xz")
```
mxcl, line 7: This is a 30MB file, pick something less than 100B pls.

Also, this is fragile. We should not rely on the Internet for testing. Really you should spin up a local web server using a tea-script and test against that.

src/hooks/useDownload.ts
kunalgoyal9: Sure... I will. Thanks for the clarification

tests/unit/cache.test.ts
mxcl: the work required for that test infrastructure is pretty high tho. However tests failing due to flakey network conditions is the worst.

However, it will pay off. We can use it to test just about everything.

jhheider: It might be as simple as `tea -X npx http-server test.fixtures/ -p 2051`

Caveats:
- 2051 = the ordinals for T(20) E(5) A(1)
- 2051 is `epnsdp`, which a quick google search might be an archaic dns thing, and used by some commercial backup solutions in their products (the port, not the service)
- the `tea -X npx` command will *also* flake if there's internet issues, but it allows a single install call before any number of http tests

src/hooks/useDownload.ts
kunalgoyal9: Fixed

README.md
```diff
@@ -799,10 +799,10 @@ export TMPDIR=${TMPDIR:-/tmp}
 deno test \
  --allow-net \
  --allow-read \
- --allow-env=SRCROOT,GITHUB_TOKEN,TMPDIR,TEA_PREFIX \
+ --allow-env=SRCROOT,GITHUB_TOKEN,TMPDIR,TEA_PREFIX,MUGGLE,MAGIC,JSON,NUMPTY \
  --allow-run \
  --import-map=$SRCROOT/import-map.json \
- --allow-write="$TMPDIR" \
+ --allow-write \
```
mxcl, line 805: why?

src/hooks/useDownload.ts
```diff
@@ -39,14 +39,21 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
...
   const { numpty } = useFlags()
   dst ??= hash().join(src.path().basename())
   if (src.protocol === "file:") throw new Error()
 
-  if (!ephemeral && mtime_entry().isFile() && dst.isReadableFile()) {
+  if (!ephemeral && dst.isReadableFile()) {
+    
     headers ??= {}
-    headers["If-Modified-Since"] = await mtime_entry().read()
+    if(etag_entry().isFile()){
```
mxcl, line 51: our code format is space before the parameter after the `if`

tests/unit/cache.test.ts
```diff
@@ -0,0 +1,24 @@
+import { assert } from "deno/testing/asserts.ts"
+import { useDownload } from "hooks"
+import { sandbox } from "../utils.ts"
+
+Deno.test("etag-mtime-check",async () => {
+  await sandbox(async tmpdir => {
+    const src = new URL("https://dist.tea.xyz/ijg.org/versions.txt")
+    await useDownload().download({src, dst: tmpdir.join("test.tar.xz")})
```
mxcl, line 8: the `dst` is wrong. Just remove it.

tests/unit/cache.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    const mtimePath = await useDownload().hash_key(src).join("mtime")
+    const etagPath = await useDownload().hash_key(src).join("etag")
```
mxcl, line 11: so we're testing an implementation detail of how useDownload() works?

Basically this test verifies useDownload() but it doesn't really test that etags work.

mxcl (changes requested):
Approved after the mods

src/hooks/useDownload.ts
```diff
@@ -77,9 +84,11 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
       dst.parent().mkpath()
       await body(reader, f, sz)
 
-      //TODO etags too
       const text = rsp.headers.get("Last-Modified")
+      const etag = rsp.headers.get("etag")
+
       if (text) mtime_entry().write({ text, force: true })
+      if (etag) etag_entry().write({text: etag, force: true})
```
jhheider, line 91: Supremely minor: I'd use consistent spacing on object notation on parallel lines.

I know our codebase as it stands would currently fail, but do you still like to "see how [you] feel" w.r.t. object notation spacing in TS, @mxcl, or should we adopt rigid guidelines? (I think we could also make this happen via eslint/prettier/etc rules and auto-application. This might prevent me from introducing semi-colons (though I don't believe I have done this in some time), and we might be able to catch imports which don't use the import map, though that might require some custom logic.)

README.md
kunalgoyal9: Because the test I have created need write access to `~/.tea/tea.xyz/var/www/https/dist.tea.xyz/` directory for writing mtime and etag files

tests/unit/cache.test.ts
kunalgoyal9: if there is no `dst` specified it will download it to `tea` directory i.e. `~/.tea/tea.xyz/var/www/https/dist.tea.xyz`

src/hooks/useDownload.ts
kunalgoyal9: Fixed

tests/unit/cache.test.ts
mxcl: well I guess you're right in that you shouldn't do that, but you are currently downloading to a file called `test.tar.xz` which is flagrantly wrong. Anyone reading this code in the future will wonder what the fuck you were thinking.

README.md
mxcl: So set the new `--allow-write` to `--allow-write="$TMPDIR:$TEA_PREFIX/tea.xyz/var/www".

We don’t allow EVERYTHING just because one new thing needs more, you keep things as restrictive as possible.

mxcl: Frankly we should change TEA_PREFIX for the tests anyway.

tests/unit/cache.test.ts
```diff
@@ -0,0 +1,24 @@
+import { assert } from "deno/testing/asserts.ts"
+import { useDownload } from "hooks"
+import { sandbox } from "../utils.ts"
+
+Deno.test("etag-mtime-check",async () => {
+  await sandbox(async tmpdir => {
+    const src = new URL("https://dist.tea.xyz/ijg.org/versions.txt")
+    await useDownload().download({src, dst: tmpdir.join("version.txt")})
```
mxcl, line 8: `versions.txt` not `version.txt`. This is sloppy.

README.md
jhheider: Comma, I think, not colon. You've spent too much time on the POSIX work.

kunalgoyal9: Fixed it, It is working on local but tests are getting failed. 

Seems like this is another issue with `$SRCROOT`. Tests that were working fine earlier are now failing :(

src/hooks/useDownload.ts
```diff
@@ -77,9 +84,11 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
       dst.parent().mkpath()
       await body(reader, f, sz)
 
-      //TODO etags too
       const text = rsp.headers.get("Last-Modified")
+      const etag = rsp.headers.get("etag")
```
mxcl, line 88: Case sensitivity doesn't matter, but we capitalized it like the docs do on the previous line so consistency would have been `ETag` per the docs for ETags.

mxcl:
In the end, this was good work, thanks @kunalgoyal9

PR #151, opened 2022-11-14
markdownlint README.md (deeply optional)

jhheider:
I don't know if this better or not, but it makes my VSCode markdownlint plugin stop complaining about these mostly stylistic choices


mxcl:
Surprisingly the lint is excellent.

PR #155, opened 2022-11-16
More careful code for all of `-x` mode

mxcl:
not merging until coverage on these pieces is higher.

Shame we have no automated code-coverage :(


src/app.exec.ts
```diff
@@ -197,7 +197,7 @@ async function repl(installations: Installation[], env: Record<string, string>)
   const shell = Deno.env.get("SHELL")?.trim() || "/bin/sh"
   const cmd = [shell, '-i'] // interactive
 
-  //TODO other shells pls #help-wanted
+  //TODO other shells pls #help-wlanted
```
jhheider, line 200: um?

mxcl:
k well, writing the tests is daunting. Needs a lot of infra and I have to move on to other things for momentum reasons.

yolo?

jhheider:
Yes, I think until we have a much broader base, there will be some yoloing of major overhauls. Since they markedly increase quality, I believe we will be forgiven for minor hits to stability along the way.

_for now._

src/app.exec.ts
mxcl: lol

PR #160, opened 2022-11-17
Fuzzy check logic implemented using levenshtein distance algorithm.

chiraggandhi123:


src/hooks/usePantry.ts
```diff
@@ -181,9 +181,52 @@ function entry({ project }: { project: string }): Entry {
     const versions = dir.join("versions.txt")
     return { dir, yml, versions }
   }
-  throw new TeaError('not-found: pantry: package.yml', {project})
-}
+  //check for fuzzy string
+  const closestPkg = getClosestPackageSuggestion(project, prefix);
```
jhheider, line 185: Probably simpler to just `await` this line.

src/hooks/usePantry.ts
```diff
@@ -181,9 +181,52 @@ function entry({ project }: { project: string }): Entry {
...
+    console.log("did you mean", {data}, "?")
+  });
+  throw new TeaError('not-found: pantry: package.yml', {project}, )
 
+}
+async function getClosestPackageSuggestion(orgPkg:string, projectPath:Path): Promise<string>{
+  let closestPkg = '';
+  let minDistance = Infinity;
+  const pkgList = [];
+  for await (const data of projectPath.ls()){
```
jhheider, line 196: I _think_ `ls()` from `usePantry.ls.ts` will work here.

jhheider (commented):
This is really cool. I think `getClosestPackageSelection` is definitely something that can be used to improve UX. My only concern, since it involves walking the pantry(ies), is that it will eventually be _sloooooow_. That said, we a) walk the pantries in other places, and b) this is alternative to throwing an error, so it'll usually be the last step before a failure. This might be the sort of thing to gate behind an increased Verbosity level, so as to not accidentally increase runtime of normal performance.

src/hooks/usePantry.ts
```diff
@@ -181,9 +181,52 @@ function entry({ project }: { project: string }): Entry {
...
+    pkgList.push(name);
+  }
+  pkgList.forEach(pkgName=>{
+    const number = levenshteinDistance(pkgName, orgPkg);
+    if(number<minDistance){
+      minDistance = number;
+      closestPkg = pkgName;
+    }
+  })
+  return closestPkg;
```
kunalgoyal9, line 207: Instead of always returning `closestPkg`, Can we use threshold here?
Something like if we don't find a package related to the ones we support.

like `random12323`

chiraggandhi123: Sure, The algorithm works in such a way, if you need to convert string 1 to string 2 then it tells us the minimum number of operations required using only three operations.
1. add character
2. modify character
3. remove character

So if the value of operations is greater than 2 or 3 we can avoid showing the suggestion. It won't be a typo.

src/hooks/usePantry.ts
chiraggandhi123: Yes, it works. Thanks.

src/hooks/usePantry.ts
chiraggandhi123: I tried using await here, but for it to the work the `entry` function must be wrapped in async but it gives some error. I will try it out though.

src/hooks/usePantry.ts
kunalgoyal9: oh Ya That could work

src/hooks/usePantry.ts
```diff
@@ -181,12 +181,50 @@ function entry({ project }: { project: string }): Entry {
     const versions = dir.join("versions.txt")
     return { dir, yml, versions }
   }
-  throw new TeaError('not-found: pantry: package.yml', {project})
+  
+  const closestPkg = await getClosestPackageSuggestion(project);
```
mxcl, line 185: let’s not do this here, instead do it in the error handler in `app.ts`. This way we don’t have to change everything from sync to async. diff will be smaller, and the error handler is really the right place to do this. Once this code is librified it really should be up to the caller whether or not they want to do this.

mxcl: I guess actually it'll be a little tricky with the current infra, so I'll do it after the merge.

Let’s get tests passing and great! Thanks so much for this contribution.

mxcl (approved):

mxcl:
Notably could be better:

```sh
$ tea run +wget    
error: not-found in pantry: wget/package.yml (spilt-tea-101)

did you mean `go.dev`?
```

would ideally suggest `gnu.org/wget`

chiraggandhi123:
> Notably could be better:
> 
> ```shell
> $ tea run +wget    
> error: not-found in pantry: wget/package.yml (spilt-tea-101)
> 
> did you mean `go.dev`?
> ```
> 
> would ideally suggest `gnu.org/wget`

Yeah, this can be achieved. we need to slightly tweek the algorithm. I will do it.

PR #173, opened 2022-11-19
support bash for tea shell

jonchang:
Add a few cases to support `bash` and `sh`. We can't pass `--norc` and `--noprofile` to `sh` since those are GNU-isms.


mxcl:
You diamond you.

PR #175, opened 2022-11-20
fix(dump): update fish command-not-found

branchvincent:
Thanks for tea, loving it so far! Here's a few quick fixes for fish I encountered:

- Command not found function is called [fish_command_not_found](https://fishshell.com/docs/current/cmds/fish_command_not_found.html)
- Lists are [1-indexed](https://fishshell.com/docs/current/language.html?highlight=List+indices+start#lists)
- Unquote `"$argv"` to prevent `tea` from intrepreting all arguments as the command (currently fails with `tea: command not found: [ "node --version" ]`)

Also added an extra newline for formatting.

Before:

```fish
function command_not_found_handler
  switch "$argv[0]"    case wget; tea '+gnu.org/wget' "$argv"
    case node; tea '+nodejs.org' "$argv"
  end
end
```

After:

```fish
function fish_command_not_found
  switch "$argv[1]"
    case wget; tea '+gnu.org/wget' $argv
    case node; tea '+nodejs.org' $argv
  end
end
```

### DIFF
--- a/src/app.dump.ts
+++ b/src/app.dump.ts
@@ -170,13 +170,13 @@ export default async function dump(args: Args) {
       break
     case 'fish':
       rv = undent`
-        function command_not_found_handler
-          switch "$argv[0]"`
+        function fish_command_not_found
+          switch "$argv[1]"\n\n`;
       for (const uninstalled of pending) {
         const exes = await pantry.getProvides(uninstalled)
         if (exes.length) {
           const cmds = exes.join(" ")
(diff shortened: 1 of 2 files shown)


jhheider (approved):
Looks good. Needs a rebate to current `main` for a couple of cosmetic changes to pick up.

mxcl:
> Needs a rebase

just squashed it. Was my bad the other commit showed up.

PR #176, opened 2022-11-20
chore: add the retried package to the import map

rossilor95:
This is needed to implement Exponential Backoff to S3 uploads in `pantry.core/scripts/upload.ts`

Issue reference: https://github.com/teaxyz/pantry.core/issues/32


import-map.json
```diff
@@ -1,19 +1,20 @@
 {
   "imports": {
-    "path": "./src/vendor/Path.ts",
```
mxcl, line 3: we'd prefer not re-ordering such things in an unrelated PR, but CI passed and whatever. :D

src/hooks/useDownload.ts
```diff
@@ -85,9 +85,9 @@ async function internal<T>({ src, dst, headers, ephemeral, logger }: DownloadOpt
       await body(reader, f, sz)
 
       const text = rsp.headers.get("Last-Modified")
-      const etag = rsp.headers.get("etag")
+      const etag = rsp.headers.get("ETag")
```
mxcl, line 88: I think this got in the PR because I force pushed. My bad. I gotta remember you can't do that after a certain point.

import-map.json
rossilor95: No problem! I'll restore the previous order 😄

PR #177, opened 2022-11-20
Slight change to algortihm 

chiraggandhi123:
It will give good suggestions of closestPackage


src/hooks/usePantry.ts
```diff
@@ -194,6 +194,7 @@ async function getClosestPackageSuggestion(orgPkg: string): Promise<string> {
     pkgList.push(project)
   }
   for (const pkgName of pkgList) {
+    if(pkgName.includes(orgPkg)) return pkgName;
```
mxcl, line 197: seems a little overkeen, but we'll see how it goes. Approved and thanks.

chiraggandhi123: Your welcome :smile:

PR #190, opened 2022-11-22
Tea foo regex change

magnusviri:
This is related to https://github.com/teaxyz/cli/issues/167. I'm trying to decide how to add another check and I'm pretty sure that the regex needs to be modified for the tea_foo test. Without this change, match[1] would be "path/tea" if the path were "/opt/tea_path/tea". I don't know why anyone would do that, but it's possible.

Also, this pull request is my way of getting my feet wet.

### DIFF
--- a/src/hooks/useFlags.ts
+++ b/src/hooks/useFlags.ts
@@ -68,7 +68,7 @@ export function useArgs(args: string[], arg0: string): [Args, Flags & Convenienc
   if (/(.+\/|^)tea_([^\/]+)$/.test(arg0)) {
     args = [...args]  // args is usually the immutable `Deno.args`
     //TODO apply muggle mode
-    const match = arg0.match(/tea_(.+)$/)!
+    const match = arg0.match(/tea_([^\/]+)$/)!
     args.unshift("-X", match[1])
   }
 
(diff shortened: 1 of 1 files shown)


magnusviri:
I thought I should probably test that this was really a bug, and it was.

```
./tea_curl/bla
Error: couldn’t find a pkg to provide: `curl/bla'

open a pull-request to add it to the pantry:

    https://github.com/teaxyz/pantry.extra
    at X (file:///Users/u0076374/Dropbox/Github/Tea/cli/src/app.X.ts:26:15)
    at async file:///Users/u0076374/Dropbox/Github/Tea/cli/src/app.ts:35:9
error: not-found: tea -X: arg0 (spilt-tea-001)
couldn’t find a pkg to provide: `curl/bla'

open a pull-request to add it to the pantry:

    https://github.com/teaxyz/pantry.extra
```

And this change causes the usage to show, which is the right behavior.

```
./tea_curl/bla
usage:
  tea [-xdX] [flags] [+package~x.y] [file|URL|target|cmd|interpreter] -- [arg…]
...
```

magnusviri:
I'm just going to close this because it's in the other pull request.

PR #193, opened 2022-11-22
read the shebang and magically hydrate deps

mxcl:


mxcl:
k, just realized that we should also support `#!/usr/bin/env foo`

mxcl:
Other considerations:

```sh
$ curl -O https://gist.githubusercontent.com/i0bj/2b3afbe07a44179250474b5f36e7bd9b/raw/colors.go
$ mv colors.go colors

$ tea ./colors
tea: installed: ~/.tea/go.dev/v1.19.2
error: permission denied (os error 13)
```

Because we just ask `sh` to do it this happens. Is this *correct*? Should we require the user to make the script executable first?

jonchang:
I think it should work without marking it executable. What we're trying to replicate here, I think, is when you have some script and you want to call it like `bash myscript` (instead of `./myscript`, which _will_ read the shebang)

I tried this diff and it seems to work:

```diff
diff --git a/src/app.exec.ts b/src/app.exec.ts
index 657b86a..696d84d 100644
--- a/src/app.exec.ts
+++ b/src/app.exec.ts
@@ -126,11 +126,12 @@ async function exec(ass: RV1, pkgs: PackageSpecification[], opts: {env: boolean}
     if (yaml) pkgs.push(...yaml.pkgs)
 
     if (magic) {
-      const found = await async_flatmap(extract_shebang(ass.path), which)
+      const shebang = await extract_shebang(ass.path)
+      const found = await async_flatmap(shebang, which)
 
       if (found) {
         pkgs.unshift(found)
-        cmd.unshift("/bin/sh")  // `run` is rudimentary so you gotta give it da `sh`
+        cmd.unshift(shebang)
       } else {
         const unshift = (project: string, ...new_args: string[]) => {
           if (!yaml?.pkgs.length) {
```

but I don't think it properly handles e.g. switches in the shebang. In any case I think it makes more sense to try to exec the script instead of shelling out to `/bin/sh`.

src/app.exec.ts
```diff
@@ -50,10 +50,12 @@ export default async function(opts: Args) {
 async function refine(ass: RV2): Promise<RV1> {
   const { magic } = useFlags()
 
+  console.log({ ass })
```
jhheider, line 53: Don't forget in final merge.

src/app.exec.ts
```diff
@@ -119,33 +121,45 @@ async function exec(ass: RV1, pkgs: PackageSpecification[], opts: {env: boolean}
...
-        case ".rb": unshift("ruby-lang.org", "ruby"); break
-      }
     }
 
     const { env } = await install(pkgs)
 
     supp(env, blueprint)
     if (yaml?.env) Object.assign(env, yaml.env)  // explicit YAML-FM env takes precedence
 
+    console.log({ cmd, env})
```
jhheider, line 161: And this one.

src/app.exec.ts
```diff
@@ -166,6 +180,43 @@ async function exec(ass: RV1, pkgs: PackageSpecification[], opts: {env: boolean}
 
 ////
 
+import {readLines} from "deno/io/buffer.ts"
+
+async function extract_shebang(path: Path) {
+  const f = await Deno.open(path.string, { read: true })
+  const line = (await readLines(f).next()).value
+  const shebang = line.match(/^\s*#!(\S+)$/)?.[1]
```
jhheider, line 188: This might be overly accepting. I bet a lot of tools don't accept leading spaces. but that shouldn't break anything.

This looks like it won't handle /usr/bin/env ruby, for example. That's probably ok, but just noting it.

src/app.exec.ts
```diff
@@ -166,6 +180,43 @@ async function exec(ass: RV1, pkgs: PackageSpecification[], opts: {env: boolean}
...
+    return { project: 'python.org', constraint: new semver.Range("3") }
+  }
+
+  const pantry = usePantry()
+  let found: { project: string } | undefined
+
+  for await (const entry of pantry.ls()) {
+    if (found) break
+    pantry.getProvides(entry).then(provides => {
+      if (!found && provides.includes(arg0)) {
```
jhheider, line 209: If we strip the bin/ prefix, should we drop that in the pantry?

src/utils/error.ts
```diff
@@ -51,7 +51,8 @@ export default class TeaError extends Error {
 
         open a pull-request to add it to the pantry:
 
-            https://github.com/teaxyz/pantry.extra
+            https://github.com/teaxyz/pantry.zero#contributing
```
jhheider, line 54: Might want to modify line 52 to include `. see` before the colon, so we don't get new package PRs to pantry.zero.

jhheider (approved):
Looks good and appropriate.

jhheider:
> Because we just ask `sh` to do it this happens. Is this _correct_? Should we require the user to make the script executable first?

I think the answer is no. However, `sh ./foo` works regardless of execute (but obviously not read) perms.

src/app.exec.ts
mxcl: I added `env` support.

I am accepting leading spaces because (amazingly) I discovered most interpreters also do this.

src/app.exec.ts
mxcl: maybe. I added it to pantry so in future we *could* eg. provide other UNIX dirs eg. `share/scripts/foo` might well be something, or ofc `sbin/foo`

mxcl:
Thanks both, good review. I'll make it so it doesn’t need to have the executable bit set.

One remaining thing that I don’t know about is how “shebangs” work exactly. Because eg. `deno` will run a script without `run` being specified, but only when running as a shebang. There must be something set in the environment be `sh` that allows `deno` (et al) to know this.

This only affects our ability to run scripts without special casing eg. `deno` (and we're moving that info to the pantry anyway, so maybe it doesn't matter).

mxcl:
Needs tests, but so does *everything*. We need this to fix a few one-liners we specify as demos so will merge.

PR #204, opened 2022-11-26
Adds tea.yml support. closes #201

### LINKED ISSUE #201: Allow a dedicated dependency file for magic (not package.json or readme)
First off, I'd like to say that I love the concept of this project and I really like the idea of a readme outlining (and defining) a projects dependencies. **but hear me out!**

I have put together a couple of reasons why I think restricting the dependencies to existing files/formats hurts the usage of this project more than it aids it. Particularly when working in a team. (I understand that motivating developers to use the readme is likely the main reason this isn't already a feature)

- By only allowing a magic environment to be defined in a file that is likely to already exist in a project makes tea harder to sell to colleagues/peers/other devs. Different people have different opinions on what a readme should/shouldn't contain, likewise for a package.json. By having a dedicated config in source control, devs can learn the advantages of tea at their own pace. Once they see how great it is, convincing them that maybe it's time to move this to the readme becomes an easier sell. 

- In a similar vein, not all teams are going to want to adopt this tool, if there were a dedicated config file one could add it to the .gitignore, use tea to their hearts desire without impacting the project or team directly and still reap most of the benefits of the tool. This currently doesn't seem possible in a nodejs project with an existing readme.

- While markdown is great, there are other great markup formats such as asciidoc that all have their pros and cons. Forcing a particular format goes against the beauty of having different tools to suit different needs. Especially if a project has an extensive readme already written up in a different format. Personally I love markdown, it's my preferred syntax but I think it's wrong to ignore where it falls short and why it may not always be the best fit for someone.

- *But you can also use package.json!* Sure but why? Package.json is a project specific file that exists to outline the metadata about a node.js project, why should this dictate the environmental config for what could be a project much larger than just the node component? Furthermore the mere existence of a package.json implies a lot about a project. It's out of place and potentially misleading to exist in a golang project for example. This leads on to my next point.

- By glancing at a directory we can typically make relatively safe assumptions. Eg. 'Ah so there's a package.json, a tsconfig.json and an .env file, it's very likely this is a node project, written in typescript with some environmental config'. This goes for all kinds of tooling and their respective configs. By having, lets say, a `teaxyz.json`, it's immediately obvious that there's some environmental config I can use to get up and running. This is almost as powerful as having the deps right there in the readme.

These criticisms are with the best of intentions, I truly believe this project could change the face of development. Apologies if they come across too harshly, I have a tendency to over explain things.

I'm definitely open to hearing why my outlined points are all wrong!

If this is something that the maintainers of the project would be willing to accept then I'll gladly do the work and submit a PR.

jakehl:
*closes #201*
### What
This PR adds support for a `tea.yml` and `.github/tea.yml` config file to drive magic functionality. The schema for the yml file copies that of the existing package.json support. 

Eg.
- tea.yml
```yml
version: 1.0.1
tea:
  dependencies:
    nodejs.org: 19
    go.dev: 1.19
```
- package.json
```json
{
  "version": "1.0.1",
  "dependencies": {
    "nodejs.org": 19,
    "go.dev": "1.19"
  }
}
```

### TODO:
- [x] Add tea.yml loader
- [x] Refactor json logic to support yaml and share schema
- [x] Move common requirements logic/types to `sr/common/` directory 
- [x] Investigate existing behavior whereby a package.json (and now a tea.yml) will prevent markdown scripts from running.
   - I would consider this out of the scope of this ticket. I'll create an issue.
   running `touch package.json` in the root of the cli repo will break commands such as `tea install-self`
- [x] Understand what the version in the config is for
   - It's the version of the project using tea, semver string. got it!
- [x] Remove junky console logs
- [ ] Update docs
- [ ] .github/tea.yml support
- [ ] UX - Do we want to let the user know which file has been loaded? 
- [ ] fix failing tests
- [ ] add new tests?

### DIFF
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -1,0 +1,3 @@
+{
+  "importMap": "import-map.json"
+}
--- a/import-map.json
+++ b/import-map.json
@@ -6,6 +6,7 @@
     "hooks/": "./src/hooks/",
     "prefab": "./src/prefab/index.ts",
     "prefab/": "./src/prefab/",
+    "common/": "./src/common/",
     "deno/": "https://deno.land/std@0.156.0/",
     "semver": "./src/utils/semver.ts",
     "utils": "./src/utils/index.ts",
--- a/src/app.help.ts
+++ b/src/app.help.ts
@@ -14,6 +14,7 @@ export default async function help() {
   if (!verbose) {
     //        10|       20|       30|       40|       50|       60|       70| |     80|
     await print(undent`
+      // TODO JL REMOVE !!!!!! DEV MODE
       usage:
         tea [-xdX] [flags] [+package~x.y] [file|URL|target|cmd|interpreter] -- [arg…]
 
--- a/src/app.help.ts
+++ b/src/app.help.ts
@@ -14,7 +14,6 @@ export default async function help() {
   if (!verbose) {
     //        10|       20|       30|       40|       50|       60|       70| |     80|
     await print(undent`
-      // TODO JL REMOVE !!!!!! DEV MODE
       usage:
         tea [-xdX] [flags] [+package~x.y] [file|URL|target|cmd|interpreter] -- [arg…]
 
--- a/src/common/requirementsCandidate.ts
+++ b/src/common/requirementsCandidate.ts
@@ -1,0 +1,47 @@
+export const markdown_extensions = [
+  "md",
+  "mkd",
+  "mdwn",
+  "mdown",
+  "mdtxt",
+  "mdtext",
+  "markdown",
+  "text",
+  "md.txt",
+] as const;
+type MarkdownExtension = typeof markdown_extensions[number];
+
+const yaml_extensions = ["yml", "yaml"];
+type YamlExtension = typeof yaml_extensions[number];
+
+export type RequirementsCandidate =
+  | "package.json"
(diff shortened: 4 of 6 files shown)


jakehl:
Sounds good to me! I'm away for Xmas at the moment but I'll set time aside for this in a week or so.

mxcl:
`tea.yml` or `tea.yaml` support officially dropped in 0.23.

dudicoco:
@mxcl can you please elaborate why support for `tea.yaml` was dropped?
It's still in the docs: https://docs.tea.xyz/features/developer-environments#supported-files

How can we manage versions declaratively without it? Version files such as `.node-version` do not exist for all packages and even if they do tea doesn't support all of them.

jhheider:
I believe he means it "dropped", was released, rather than "was dropped", discarded. Indeed, `tea.yaml` is present and functional in [`teaxyz/pantry`](https://github.com/teaxyz/pantry/blob/main/tea.yaml).

mxcl:
Yes sorry to be confusing. I meant it went live. We will be keeping this feature forever!

dudicoco:
Thanks @jhheider @mxcl.

It seems that the `tea.yaml` parameters are not documented anywhere.

Also, is it possible to install all of the dependencies specified in the file like with `brew bundle`?

PR #207, opened 2022-11-28
Add runtime environment variables configuration

zegl:
Allow specifying env vars to be set at runtime in pantry configs.

Example:

```yaml
runtime:
  env:
    DENO_NO_UPDATE_CHECK: "true"
```

Updates #26

### DIFF
--- a/src/hooks/usePantry.ts
+++ b/src/hooks/usePantry.ts
@@ -31,6 +31,7 @@ export default function usePantry() {
     getProvides,
     getYAML,
     getInterpreter,
+    getRuntimeEnvironment,
     resolve,
     ls,
     prefix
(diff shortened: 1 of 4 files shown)


src/hooks/usePantry.ts
```diff
@@ -185,6 +186,11 @@ const getInterpreter = async (_extension: string): Promise<Interpreter | undefin
   return undefined
 }
 
+const getRuntimeEnvironment = async (pkg: {project: string}): Promise<Record<string, string>> => {
+  const yml = await entry(pkg).yml()
+  return yml["runtime"]?.["env"] ?? {}
```
mxcl, line 191: Should validate it's a `PlainObject`. I'll add in subsequent patch.

mxcl (approved):

mxcl:
Tests fail due to revlock

PR #208, opened 2022-11-28
Use sandbox-exec on macOS to disable writes to TEA_PREFIX

### LINKED ISSUE #77: use `sandbox-exec` on macOS to restrict writes to TEA_PREFIX

zegl:
Runs the repl or commands inside a sandboxed environment.

The sandbox allows everything by default, and then disables writes to all files under TEA_PREFIX.

Fixes #77

--- 

edit: ok, there's quite a few tests at the moment that _expect_ TEA_PREFIX to be writeable 😅

### DIFF
--- a/src/app.exec.ts
+++ b/src/app.exec.ts
@@ -112,7 +112,7 @@ async function exec(ass: RV1, pkgs: PackageSpecification[], opts: {env: boolean}
 
     const cmd = [arg0, ...ass.args]
 
-    await run({cmd, env, useSandboxExec: true})
+    await run({cmd, env, useSandboxExec: false})
 
   } break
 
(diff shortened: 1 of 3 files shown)


zegl:
Example:

```
$ ~/s/t/cli (sandbox-exec)> tea run -S +deno.land
tea: pantries sync’d ⎷
this is a temporary shell containing the following packages:
deno.land@1.28.2
when done type: `exit'
tea %~ touch $TEA_PREFIX/what
touch: /Users/gustav/.tea/what: Operation not permitted
tea %~ touch ~/what
tea %~
```

mxcl:
#77 actually wanted the opposite (wording was poor, my bad 😓), restricting writes *only to `TEA_PREFIX`*.

I think though I was thinking about “builds” when making that ticket, and well, we have since decided that users won't do builds. However it would still be a welcome change to @teaxyz/pantry-core’s `build.ts`.

For the REPL, I'm not sure we should sandbox that either.

open to a discussion about what *should* be sandboxed.

I think it could be *amazing* to sandbox tea-scripts.

mxcl:
I've been noodling this and it's possible this is totally the right time to add this. Where packages *must* specify what paths in their own prefix they allow writes. And we figure out the issues with the tests.

Most packages should not allow writes in $TEA_PREFIX. This is a nice feature.

PR #221, opened 2022-12-02
docs: add small documentation and navigation changes to README

mfts:


mxcl:
I think these changes are sensible, though we should work on it a bit.

One issue with them is tea requires the first header to contain the version so we can extract it. Perhaps this is bad? But it's the case and we'd need to change the code if we were going to keep this structure.

Also `tea is not a package manager` is like a thing I like :)

mxcl:
https://github.com/mfts/tea-cli/tree/docs/improvements <-- to see changes

PR #227, opened 2022-12-05
POC: put a GUI inside the CLI

getneil:
This is just a quick and very dirty POC: https://www.loom.com/share/be3b09a0e6794dbf84001574036e2641

The idea is to be able to try to install packages via a UI.

how to use: `$ tea gui` then it will open `localhost:5001`
This will add the web static files in `$HOME/.tea/tea.xyz/tmp/gui`

There are several questions that need to be answered first:
1. where to put the GUI static web files. Ideally there is no .zip file here( or maybe there is one). How to update it if there is a bug? maybe it updates the same way as the tea cli auto updates itself?
2. how to update it, maybe compare it with a remote version every time `$ tea gui` is executed
3. how to share TS types/shapes between gui and cli? maybe they need to become a mono repo?


jhheider:
I think the answer to all 3 questions is probably to add tea.xyz/gui to the sync and use symlinks to cli's import-map.json and src directories the same way the pantries do.

PR #234, opened 2022-12-06
Adding deno.config 

otherJL0:
Adding a `deno.config` file provides:

* A recognizable file for the `deno lsp` to reference when working in editors other than VSCode
* A set of tasks to build and test with all flags with `deno task build` and `deno task test`

At this time the "fmt" section of the config is set to ignore all files since the current style deviates from deno's default style.


mxcl:
Do we still need `tsconfig.json`?

otherJL0:
We don't really need `tsconfig.json` anymore, I can remove it and update the PR. And yeah we could update the README with the new task commands.

lino-levan:
Any particular reason why this is a `.json` and not a `.jsonc`? I don't think it matters yet but if we ever want to add comments to it I think it would be valuable to have.

jhheider:
> At this time the "fmt" section of the config is set to ignore all files since the current style deviates from deno's default style.

A better question is: "should we codify our deviations and enable formatting?"

otherJL0:
`deno-config.json` is now `deno-config.jsonc` to allow for future comments. The README has been updated to use the more succinct `deno task` syntax for local development

PR #235, opened 2022-12-06
Don’t unquarantine

### LINKED ISSUE #157: permissions on tea-installed executables trigger corporate security monitoring
The use of `xattr com.apple.quarantine` to set permissions on tea-installed executables sets off reports with my company's global security operations center.

https://github.com/teaxyz/cli/blob/main/src/prefab/install.ts#L100

I ran the example command:
```tea +gnu.org/wget wget -qO- tea.xyz/white-paper```

And promptly got a warning from my global security team about permissions escalation:
```
File name: find
File path: /usr/bin/find
Command line: find /Users/amclin/.tea/openssl.org/v1.1.118 -xattrname com.apple.quarantine -perm -0200 -exec xattr -d com.apple.quarantine {} ;
```

They are using CrowdStrike to monitor this, but I have no direct way of testing allow/deny lists with them or CrowdStrike.

I am unclear if the flag happened because of the use of `-d com.apple.quarantine`, elevating the permission level to `0200`, or the masking of the command within `find`, but I suspect it is likely the removal of the `com.apple.quarantine` flag that set it off as I've never run into this alert from normal manual usage of `chmod` or by manually authorizing executables via the control panel's security settings.

mxcl:
Fixes #157

### DIFF
--- a/src/prefab/install.ts
+++ b/src/prefab/install.ts
@@ -65,8 +65,6 @@ export default async function install(pkg: Package, logger?: Logger): Promise<In
 
     const install = await cellar.resolve(pkg)
 
-    await unquarantine(install)
-
     const str = [
       gray(usePrefix().prettyString()),
       install.pkg.project,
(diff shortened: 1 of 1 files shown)


mxcl:
@jhheider I blew away my `~/.tea` installation to test this and so far nothing is broken. I tested `deno` first.

PR #236, opened 2022-12-06
fix: Outputting color codes on non-tty

lino-levan:
I don't know if this is the best way to do it, but it certainly is *a way* to do it. I think it is somewhat elegant that it does not need to modify the behavior of any other files.

ref: #3


lino-levan:
Closing this and reopening another PR with a signed commit

PR #237, opened 2022-12-06
fix: Outputting color codes on non-tty

lino-levan:
I don't know if this is the best way to do it, but it certainly is *a* way to do it. I think it is somewhat elegant that it does not need to modify the behavior of any other files.

ref: https://github.com/teaxyz/cli/issues/3


mxcl:
Great thanks. I think indeed this is the only way to do it such that we are sure we won't accidentally output color codes with future code.

PR #244, opened 2022-12-08
Fixed Windows path issue

calvinyee:
Fixed Path.ts to get pass first Tea Path error (1- changing slashes for windows and 2- using another deno function that has to do with symlink)

### DIFF
--- a/src/vendor/Path.ts
+++ b/src/vendor/Path.ts
@@ -51,14 +51,14 @@ export default class Path {
       case "darwin":
         if (!input || input[0] != '/') {
           throw new Error(`invalid absolute path: ${input}`)
-        } else {	    
+        } else {
           this.string = sys.normalize(input)	    
         }
         break
-      case "windows":               
+      case "windows":                   
        if (!input || !(/^(["[a-zA-Z]:[\\\/].*|[\\\/].*)$/.test(input))) {          
           throw new Error(`invalid absolute path: ${input}`)
-        } else {	          
+        } else {
           this.string = slash(sys.normalize(input))	    
         }
         break
@@ -85,7 +85,8 @@ export default class Path {
     */
   readlink(): Path {
     try {
-      const output = Deno.build.os == "windows" ? Deno.realPathSync(this.string) : Deno.readLinkSync(this.string)
+      // with readLinkSync we get on windows the following error: The file or directory is not a reparsed point      
+      const output = Deno.build.os == "windows" ? Deno.realPathSync(this.string) : Deno.readLinkSync(this.string)      
       return this.parent().join(output)
     } catch (err) {
       const code = err.code
--- a/src/vendor/Path.ts
+++ b/src/vendor/Path.ts
@@ -55,10 +55,10 @@ export default class Path {
           this.string = sys.normalize(input)	    
         }
         break
-      case "windows":
-       if (!input) {
+      case "windows":               
+       if (!input || !(/^(["[a-zA-Z]:[\\\/].*|[\\\/].*)$/.test(input))) {          
           throw new Error(`invalid absolute path: ${input}`)
-        } else {	    
+        } else {	          
           this.string = slash(sys.normalize(input))	    
         }
         break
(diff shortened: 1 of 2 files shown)


src/vendor/Path.ts
```diff
@@ -3,6 +3,7 @@ import * as sys from "deno/path/mod.ts"
 import * as fs from "deno/fs/mod.ts"
 import { PlainObject } from "is_what"
 import {readLines} from "deno/io/buffer.ts"
+import slash from "https://deno.land/x/slash@v0.3.0/mod.ts";
```
jhheider, line 6: For version control purposes, add `"slash": "https://deno.land/x/slash@v0.3.0/mod.ts",` to `import-map.json` and use `import slash from "slash"`.

src/vendor/Path.ts
```diff
@@ -38,15 +39,28 @@ export default class Path {
...
+    switch (Deno.build.os) {
+      case "linux":
+      case "darwin":
+        if (!input || input[0] != '/') {
+          throw new Error(`invalid absolute path: ${input}`)
+        } else {	    
+          this.string = sys.normalize(input)	    
+        }
+      case "windows":
+       if (!input) {
```
jhheider, line 58: This is harder. Windows absolute paths can be "[letter]:\" or "\\.*", [I believe](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN)

src/vendor/Path.ts
calvinyee: Thanks for the comment.  I committed this change.

src/vendor/Path.ts
calvinyee: This is harder.  Need to figure out all the possible cases ?

jhheider: Need at least a little bit of trapping for invalid absolute paths. So, I think matching either [a-zA-Z]:/ or // will do most of it.

mxcl:
For this I would like the following (unless impossible or stupid, LMK)

* Internally use forward slash as path separator
* convert for `.pretttyString()`

I believe that all APIs will accept forward slashes, even Win32. But could be wrong.

calvinyee:
> For this I would like the following (unless impossible or stupid, LMK)
> 
> * Internally use forward slash as path separator
> * convert for `.pretttyString()`
> 
> I believe that all APIs will accept forward slashes, even Win32. But could be wrong.

We use forward slash internally as path separator. Slash() function does the conversion from backslash to forward slash if necessary.

src/vendor/Path.ts
```diff
@@ -55,10 +55,10 @@ export default class Path {
           this.string = sys.normalize(input)	    
         }
         break
-      case "windows":
-       if (!input) {
+      case "windows":               
+       if (!input || !(/^(["[a-zA-Z]:[\\\/].*|[\\\/].*)$/.test(input))) {          
           throw new Error(`invalid absolute path: ${input}`)
-        } else {	    
+        } else {
```
mxcl, line 61: please don’t change whitespace

src/vendor/Path.ts
```diff
@@ -38,15 +39,30 @@ export default class Path {
...
+      case "linux":
+      case "darwin":
+        if (!input || input[0] != '/') {
+          throw new Error(`invalid absolute path: ${input}`)
+        } else {	    
+          this.string = sys.normalize(input)	    
+        }
+        break
+      case "windows":               
+       if (!input || !(/^(["[a-zA-Z]:[\\\/].*|[\\\/].*)$/.test(input))) {
```
mxcl, line 59: Seems like this is not a valid regex for the domain (`.*` rather than `.+`). However it seems enough to simply check for an initial `/^[a-zA-Z]:\\/` to me.

src/vendor/Path.ts
```diff
@@ -69,7 +85,7 @@ export default class Path {
     */
   readlink(): Path {
     try {
-      const output = Deno.readLinkSync(this.string)
+      const output = Deno.build.os == "windows" ? Deno.realPathSync(this.string) : Deno.readLinkSync(this.string)
```
mxcl, line 88: please add comments explaining why unexpected things are necessary. To me, this seems wrong.

src/vendor/Path.ts
calvinyee: I tried your suggestions and they didn't work unfortunately. We can have 0 following characters and we can also have paths like / or \\. So I left my initial regex.

PR #246, opened 2022-12-09
Create CODE_OF_CONDUCT.md

pedrocruzio:


jonchang:
I suggest creating a new `.github` repository and putting these files in there, instead. This means that the relevant sections will automatically appear in all repositories in the `teaxyz` organization, rather than on a repository-by-repositoy basis.

pedrocruzio:
@jonchang good point! I will resubmit.

PR #247, opened 2022-12-09
Update issue templates

pedrocruzio:


mxcl:
If we do add this I'd rather it was actually useful rather than copy and pasted from something.

PR #248, opened 2022-12-09
Create SECURITY.md

pedrocruzio:


jonchang:
I suggest creating a new `.github` repository and putting these files in there, instead. This means that the relevant sections will automatically appear in all repositories in the `teaxyz` organization, rather than on a repository-by-repositoy basis.

jonchang:
Looks like this pull request is against the wrong repository? Should be this one: https://github.com/teaxyz/.github

mxcl:
@pedrocruzio pls re-open in the correct place

PR #250, opened 2022-12-10
Hoo

lailamree:
Amazing


mxcl:
If possible we will trace the source of this bot and ensure they never receive any token.

This is simply a waste of *my time* and disgusts me to my very soul.

timothytlewis:
They won't pass the prior validated commit checks.

PR #255, opened 2022-12-14
arbitrary digit & openssl scheme version support

mxcl:
not finished as semver-intersect needs more work

### DIFF
--- a/src/utils/hacks.ts
+++ b/src/utils/hacks.ts
@@ -28,7 +28,7 @@ export function validatePackageRequirement(input: PlainObject): PackageRequireme
   if (constraint === undefined) {
     constraint = '*'
   } else if (isNumber(constraint)) {
-    constraint = `${constraint}`
+    constraint = `~${constraint}`
   }
   if (!isString(constraint)) {
     throw new Error(`invalid constraint: ${constraint}`)
--- a/src/utils/hacks.ts
+++ b/src/utils/hacks.ts
@@ -28,7 +28,7 @@ export function validatePackageRequirement(input: PlainObject): PackageRequireme
   if (constraint === undefined) {
     constraint = '*'
   } else if (isNumber(constraint)) {
-    constraint = `~${constraint}`
+    constraint = `^${constraint}`
   }
   if (!isString(constraint)) {
     throw new Error(`invalid constraint: ${constraint}`)
--- a/tests/unit/pkgutils.test.ts
+++ b/tests/unit/pkgutils.test.ts
@@ -35,7 +35,7 @@ Deno.test("pkg.str", async test => {
   }
 
   await test.step("range of one version", () => {
-    const constraint = new Range("1.2.3")
+    const constraint = new Range("=1.2.3")
 
     out = pkg.str({
       project: "test",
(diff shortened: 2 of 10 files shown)


src/utils/hacks.ts
```diff
@@ -28,7 +28,7 @@ export function validatePackageRequirement(input: PlainObject): PackageRequireme
   if (constraint === undefined) {
     constraint = '*'
   } else if (isNumber(constraint)) {
-    constraint = `${constraint}`
+    constraint = `~${constraint}`
```
mxcl, line 31: before we were interpreting these as `^` ranges, but semver strictly defines “just numbers” to be `~` ranges

tests/unit/pkgutils.test.ts
```diff
@@ -35,13 +35,13 @@ Deno.test("pkg.str", async test => {
...
-    const constraint = new Range("1.2.3")
+    const constraint = new Range("=1.2.3")
 
     out = pkg.str({
       project: "test",
       constraint
     })
     assert(constraint.single())
-    assertEquals(out, `test@1.2.3`)
+    assertEquals(out, `test=1.2.3`)
```
mxcl, line 45: we no longer support `@`. Mostly just to stop people being confused since `npm` uses `@` to mean “a ^ range follows this“ and because `@` has many other connotations online.

src/utils/hacks.ts
```diff
@@ -28,7 +28,7 @@ export function validatePackageRequirement(input: PlainObject): PackageRequireme
   if (constraint === undefined) {
     constraint = '*'
   } else if (isNumber(constraint)) {
-    constraint = `${constraint}`
+    constraint = `^${constraint}`
```
mxcl, line 31: This is what we’ve been meaning by our usage in package.ymls.

jhheider:
Definitely should not be. This should make it more robust, but I'll double check it.

jhheider:
Kettle is locked to, currently, v0.11.10 for semver.ts, so my use of due caution means this won't break it before it's battle-hardened. But it will break the workflow trigger:
```typescript
body: JSON.stringify({
      ref: "main",
      inputs: {
        projects: projects.map(x => `${x.project}@${x.version}`).join(" ")
      }
    })
```

I think you can see that that's a single character fix, however, and _pre-fixing it_ prior to release won't break anything either.

jhheider:
And https://github.com/teaxyz/pantry.core/commits/91b6040a8889c7708b7665eff5c3f259d2081da7

mxcl:
Looking at this usage probably we should support `@` and not document it. but whatever.

mxcl:
Actually `pkg.ts.parse()` still supports @ syntax, though it is interpreted as `exactly` and maybe we should just do what you expected Jacob and everyone expects and make it interpreted as `^`

jhheider:
I always expect '@' == '='. I find the usage of '^' as a default counter-intuitive.

mxcl:
@jhheider I disagree that you expect this. You expected `@3` to be latest 3.x.y just 7 weeks ago!

jhheider:
Sorry. Was referring to your remarks about npm defaults. I was thinking specifically about fully quantified values, not @3, but @3.1.2.

mxcl:
yeah honestly maybe people expect `@3.1.2` to not become `3.1.9`. we could do that. I wonder if nom has any other quirks we should “copy”

PR #283, opened 2023-01-05
Possible fix to shell escaping env vars

jhheider:
So. I'm not positive we should do this. This is solves the escaping of env vars in the proper `bash` way (as blessed by `shellcheck`), which is to use an array sets of things. This is good.

This is going to require changing numerous formulae in both pantries which rely on the old version of settings a single sell variable. This seems really, really bad.

You can test it by building a sample `package.yml` (I called it `foo`) that looks like this:

```yaml
distributable: ~

versions:
  - 0.0.1

build:
  script: |
    echo $ARGS
    for x in "${ARGS[@]}"; do
      echo "$x"
    done
  env:
    ARGS:
      - foo
      - bar
      - "baz bat"
      - --thing="foo bar"
      - "--thing=baz bat"

test: true
```

Since people shy away from bash arrays, I think it might be smarter just to treat them directly in the script rather than make this change wholesale. But I submit it for evaluation and discussion.

### DIFF
--- a/import-map.json
+++ b/import-map.json
@@ -11,6 +11,7 @@
     "utils": "./src/utils/index.ts",
     "utils/": "./src/utils/",
     "is_what": "https://deno.land/x/is_what@v4.1.7/src/index.ts",
+    "shell_escape": "https://deno.land/x/shell_escape@1.0.0/index.ts",
     "cliffy/": "https://deno.land/x/cliffy@v0.25.2/",
     "s3": "https://deno.land/x/s3@0.5.0/mod.ts",
     "outdent": "https://deno.land/x/outdent@v0.8.0/mod.ts",
--- a/src/hooks/usePantry.ts
+++ b/src/hooks/usePantry.ts
@@ -2,6 +2,7 @@
 import { Package, PackageRequirement, Installation } from "types"
 import { host, flatmap, undent, validate_plain_obj, validate_str, validate_arr, pkg, TeaError } from "utils"
 import { isNumber, isPlainObject, isString, isArray, isPrimitive, PlainObject, isBoolean } from "is_what"
+import { multipleArguments } from 'shell_escape'
 import { validatePackageRequirement } from "utils/hacks.ts"
 import { useCellar, useGitHubAPI, usePrefix } from "hooks"
 import { ls, pantry_paths, prefix } from "./usePantry.ls.ts"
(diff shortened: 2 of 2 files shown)


jhheider:
The fact that this passes all tests only makes me more nervous.

jhheider:
The flipside of my concern about arrays, of course, is that it's the right thing to do. strings of strings (potentially of string of strings of string) each with their own escaping concern is why shells have the reputation they do already. `bash`'s solution **is** the array, so maybe we should be good boys and girls and just use it.

jhheider:
(since this is still taking up space in my brain)

hilariously, even things that look right don't seem to be based on testing. The following don't want to work:
```sh
export ARGS="thing 'foo bar' other_thing"
export ARGS="thing \"foo bar\" other_thing"
export ARGS="thing foo\ bar other_thing"
```
`bash` really _wants_ you to use arrays for this use-case.

mxcl:
We’ll try this out after v0.19.0 has been out a little bit.

mxcl:
isn't the correct way to do this, this:

```bash
export ARGS="thing ""foo bar"" other_thing"
```

?

My concern is only due to my lack of familiarity with (ba)sh arrays and what might happen as a result of this change

OTOH we can revert it quickly enough.

jhheider:
I no longer remember if I tried double quotes. I'll test that.

jhheider:
So, this doesn't seem to work:

```sh
ARGS="$ARGS --with-pkgversion=""tea GCC {{version}}"""
../configure "$ARGS"
```

But, I agree that this is a massive change in style. I don't feel good about merging it; certainly without testing.

Unpopular opinion: switching our internal shell scripts to something like `fish`, where all variables are lists by default, solves this at the expense of exposing us to unknown fish idiosyncrasies and involves much, much, much more work and testing. Maybe the build scripts should .rb, .py, .ts, or some other real language? I don't think there's any one "best" solution, but bash is by far the least surprising.

mxcl:
Could you elaborate on what doesn’t work?

I'd be fine with fish, if it was widely known, but it isn’t. We have to make choices are contributors are comfortable with if we want contributions.

mxcl:
Possibly it didn’t work because we’ve already been doing the double-double-quote expansion for months:

```typescript
function expand_env(env: PlainObject, pkg: Package, deps: Installation[]): string {
  return Object.entries(expand_env_obj(env, pkg, deps)).map(([key,value]) => {
    // weird POSIX string escaping/concat stuff
    // eg. export FOO="bar ""$baz"" bun"
    value = `"${value.trim().replace(/"/g, '""')}"`
    while (value.startsWith('""')) value = value.slice(1)  //FIXME lol better pls
    while (value.endsWith('""')) value = value.slice(0,-1) //FIXME lol better pls

    return `export ${key}=${value}`
  }).join("\n")
}
```

jhheider:
Agree that switching to fish is worse in most ways. I believe the thing that was failing was:
```yaml
script: |
  ../configure $ARGS
env:
  ARGS:
    - --version-string="tea version {{version}}"
```

Including varieties of double-quoting.

PR #292, opened 2023-01-11
v0.19.0

mxcl:
* removes exe/md
    * (maybe it'll come back as a tea/cmd)
    * tea makes it trivial to use `make` or `just`
    * exe/md is interesting, but it’s out of scope for tea
* shell magic redesigned
    * `$ npm start` will install npm and run `start` if not installed
    * developer environments made more robust
* `tea npm` installs npm if not installed
    * `-X` flag deprecated and now implicit
    * If user specifies any `+pkg.com` flags this behavior is suppressed
    * if user doesn't want this, then they presumably would have installed `tea` without magic in the first place
    * `tea foo` will call through to the system if a package providing `foo` is not found.
        * `--dry-run` is provided so users can debug what tea is doing to help with any potential confusion here
* removes flags controlling magic
    * we redefine magic to be our shell magic
    * everything else that was in the magic bucket now either has flags or was rethought
* numerous fixes and improvements
* README is greatly simplified
     * all this work makes a lot of what tea is and how it works much more digestible and intuitive
     * eg. tea no longer needs an install command, everything installs automatically
     * eg. tea no longer needs a “search for packages” feature. Just type the command you want.


src/prefab/link.ts
```diff
@@ -3,8 +3,11 @@ import { useCellar } from "hooks"
 import Path from "path"
 import SemVer, * as semver from "semver"
 import { panic } from "../utils/safe-utils.ts"
+import useFlags from "../hooks/useFlags.ts"
```
jhheider, line 7: cleanup: import-map-based paths

src/vendor/Path.ts
```diff
@@ -104,7 +104,7 @@ export default class Path {
   /// rationale: usually if you are trying to join an absolute path it is a bug in your code
   /// TODO should warn tho
   join(...components: string[]): Path {
-    const joined = components.join("/")
+    const joined = components.filter(x => x).join("/")
```
jhheider, line 107: `.filter(x => x)` differ from `.compact()` for certain truthinesses? that would be very JavaScript-y.

mxcl: I forgot we provided compact lol

mxcl: oh we can't use it since Path is meant to become its own library.

PR #300, opened 2023-01-14
Fix fish magic script on v0.19.0

### LINKED ISSUE #299: v0.19.0 magic errors out on fish
With the following line in my config.fish:

```fish
test -d "$HOME/.tea" && "$HOME/.tea/tea.xyz/v*/bin/tea" --magic --silent | source
```

I'm getting this error message when opening my term:

```
- (line 12): Missing end to balance this if statement
if ! tea true 2>&1 >/dev/null; then
^^
from sourcing file -
	called on line 12 of file ~/.config/fish/config.fish
from sourcing file ~/.config/fish/config.fish
	called during startup
source: Error while reading file '<stdin>'
```

I'd appreciate any ideas on this.

davdroman:
Fixes #299

### DIFF
--- a/src/app.magic.ts
+++ b/src/app.magic.ts
@@ -43,7 +43,7 @@ export default function(self: Path) {
       # if the user put tea in eg. /usr/local/bin then don’t pollute their PATH
       if ! tea true 2>&1 >/dev/null; then
         export PATH="${d}:$PATH"
-      fi
+      end
       `
   case "bash":
     return undent`


jhheider (approved):
Good find.

mxcl:
Sorry for this bug. We need CI for the magic somehow.

davdroman:
Does current CI replicate installation on every shell? If so, we might be able to assert whether `fish --login` returns a status code != 0.

jhheider:
It might be as simple as `for shell in fish zsh bash; do SHELL=$shell tea --magic | $shell; done`

davdroman:
@jhheider doesn't seem to do the trick... 

image

davdroman:
Oh nevermind it does, it's just the others returning 0 that override the status code. Returns 127 if I just run it for fish. 👌 

image

jhheider:
Beautiful. I think that might be at least a first-pass test that there's no errors.

PR #312, opened 2023-01-16
chore: bump deno std to latest

lino-levan:
This is needed for https://github.com/teaxyz/pantry.core/pull/102 to land.

Marking this as a draft because there's one bit of code that doesn't have a clear transition path to the current standard library (since they removed `std/hash` in favor of the webcrypto api). Currently talking to the deno folks on how to best approach this.

### DIFF
--- a/import-map.json
+++ b/import-map.json
@@ -14,6 +14,7 @@
     "cliffy/": "https://deno.land/x/cliffy@v0.25.2/",
     "s3": "https://deno.land/x/s3@0.5.0/mod.ts",
     "outdent": "https://deno.land/x/outdent@v0.8.0/mod.ts",
+    "sha256": "https://deno.land/std@0.160.0/hash/sha256.ts",
     "rimbu/": "https://deno.land/x/rimbu@0.12.3/"
   }
 }
--- a/import-map.json
+++ b/import-map.json
@@ -6,7 +6,7 @@
     "hooks/": "./src/hooks/",
     "prefab": "./src/prefab/index.ts",
     "prefab/": "./src/prefab/",
-    "deno/": "https://deno.land/std@0.172.0/",
+    "deno/": "https://deno.land/std@0.173.0/",
     "semver": "./src/utils/semver.ts",
     "utils": "./src/utils/index.ts",
     "utils/": "./src/utils/",
--- a/import-map.json
+++ b/import-map.json
@@ -6,7 +6,7 @@
     "hooks/": "./src/hooks/",
     "prefab": "./src/prefab/index.ts",
     "prefab/": "./src/prefab/",
-    "deno/": "https://deno.land/std@0.156.0/",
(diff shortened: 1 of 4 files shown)


lino-levan:
I realize that this should also (as a bonus) be a performance increase because we're moving from a javascript implementation to a native implementation of hashing.

lino-levan:
Looks like my current solution works, but it is unbelievably slow (because I'm buffering the whole file into memory before calculating the digest). I think our best bet is to just import what `std/hash/sha256` was based on (`x/sha256`) and move on until we find a better solution.

jhheider:
> Looks like my current solution works, but it is unbelievably slow (because I'm buffering the whole file into memory before calculating the digest). I think our best bet is to just import what `std/hash/sha256` was based on (`x/sha256`) and move on until we find a better solution.

@lino-levan you're amazing for worrying about hash speed in all this. Thanks for your work on this issue.

mxcl:
so yeah we did spend some time making sure we calculated the digest as we downloaded it for speed/efficiency and memory reasons. A little disappointing Deno has lost that capability (for now presumably).

We would like to stay current with the deno stdlib though for sure, but we should not regress. Being efficient is a tenant of the project.

lino-levan:
We have a few options on how to progress here:
1. Vendor the sha256 code that was removed from std directly into the cli. This is in my opinion the best option.
2. Use another module that does the same thing (but there aren't a lot of good maintained typed options out there)
3. Give up on keeping std up to date

I'm in favor of option 1 personally. It seems that the folks at deno are being receptive of the use case so it may end up back in std at some point in the near future. Thoughts @mxcl @jhheider?

lino-levan:
Actually, there's a fourth bonus option which is to have most of the project use latest std and then specifically have one import here that uses the older std:

```typescript
import { Sha256 } from "https://deno.land/std@0.160.0/hash/sha256.ts"
```

jhheider:
Yeah, it'd be _great_ not to use different versions of `std`, but this is a trick I've used before. It'd be great if https://github.com/denoland/deno/issues/12993 had taken, because then you could easily call attention to these kinds of explicit multiple-API imports, but you can't have everything.

lino-levan:
The almighty @iuioiua has volunteered to take a quick look. Hope to sanity check that I'm not missing something super obvious.

jhheider:
I'd put it in the map for uniformity's sake (and easy evaluation of external deps).

lino-levan:
I think we can always optimize / change this later. This is blocking upstream issues so I would prioritize getting this merged over having this be perfect in the first iteration.

jhheider (approved):
Looks good to me. Thanks, @lino-levan .

mxcl (approved):

mxcl:
Wonderful, thank you for your contribution @lino-levan

PR #322, opened 2023-01-19
remove SSL_CERT_FILE insert

jhheider:
To commit after merging https://github.com/teaxyz/pantry.core/pull/110

### DIFF
--- a/src/hooks/useShellEnv.ts
+++ b/src/hooks/useShellEnv.ts
@@ -14,7 +14,6 @@ export const EnvKeys = [
   'XDG_DATA_DIRS',
   'CMAKE_PREFIX_PATH',
   'DYLD_FALLBACK_LIBRARY_PATH',
-  'SSL_CERT_FILE',
   'LDFLAGS',
   'TEA_PREFIX',
   'PYTHONPATH',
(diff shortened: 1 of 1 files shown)


mxcl:
this is some pretty gnarly revlock for our users. We shouldn't remove it. Instead check if the file is there or not as well?

jhheider:
Let's see. This will break... openssl until a tea --sync is run. After that, openssl.org/package.yml will dep ca-certs, and all will be well. That _is_ worrying. We'd basically need to check if openssl.org is in the package list and ca-certs is not, to know if they had the old one. So, the question really boils down to whether this a good YOLO to keep from getting crufted up this early in the game, or if this is too critical a component to break today. I can argue it either way.

Conversely, instead of keeping this around, what if we had a startupWarnings.ts that could check for some of these version issues and give a useful warning? So, in this instance, if !openssl.org.yml.dependencies.contains(curl.se/ca-certs), console.warn('out of date pantry revlock on openssl.org:curl.se/ca-certs. run `tea --sync` to bring your pantry up to date and resolve this issue.'). Maybe.

src/hooks/useShellEnv.ts
```diff
@@ -66,13 +65,6 @@ export default async function useShellEnv({installations}: Options): Promise<Rec
       vars.ACLOCAL_PATH = compact_add(vars.ACLOCAL_PATH, installation.path.join("share/aclocal").compact()?.string)
     }
 
-    if (installation.pkg.project === 'openssl.org') {
-      const certPath = installation.path.join("ssl/cert.pem").compact()?.string
```
mxcl, line 70: Fortunately this was already defensively programmed. So we’re safe! Path.compact is undefined if the file doesn't exist.

mxcl:
We should remove this at 1.0, or at wait at least 3 months I think. Until then we're safe. Unless I'm mistaken.

src/hooks/useShellEnv.ts
jhheider: Ah, then that will work, yes. As long as it doesn't overwrite runtime.env.SSL_CERT_FILE value from curl.se/ca-certs with nothing, all is well.

PR #330, opened 2023-01-21
Add `--provides` mode

secondary-smiles:
- `--provides` allows other programs like autocomplete to detect whether a program is available

Many people user terminals that automatically color the commandline depending on whether the program is available or not
This shaky gif demonstrates that (tea breaks vhs a bit)
demo

`tea --provides [packages]` allows programs like `fish` to poll tea before highlighting commands which should lead to a better UX


mxcl:
a lovely addition, thank you.

PR #334, opened 2023-01-23
test fire teaxyz/setup:ci.yml

jhheider:
resolves https://github.com/teaxyz/setup/issues/122

If we don't want to fail if CI does (but then what?), set `wait-for-completion: false` on the `smoke-setup:` job's single step.


mxcl:
what we need I guess is for this to trigger tea/setup’s CI/CD but not to wait on it here. Then if it fails we need a slack notification or something.

But not in the current channel for pantry build failures as it's too noisy, so probs in #dev

mxcl:
Come to think of it, if it fails it should open a ticket.

Same for new-version failures

mxcl:
Approved once you merge the other end.

PR #342, opened 2023-01-26
Fix fish magic

davdroman:
It broke again. Perhaps a good idea to set up CI for it as discussed in the previous PR (#300). I'd do it myself if I had more time but I'm currently swamped with my day job.


mxcl:
ugh, jeez. I'm sorry.

I am in fact writing tests for the magic, I'll try to finish that quickly.

mxcl:
Tests in here https://github.com/teaxyz/cli/pull/343

PR #347, opened 2023-01-28
fix workflow dispatch

jhheider:


jhheider:
hm. ok, just a stupid typo.

PR #355, opened 2023-02-03
Better errors if we don’t have a pkg version

mxcl:


mxcl:
```sh
$ tea +deno.land=1.22
resolving package graph
error: not-found: pkg.version (spilt-tea-004)
we haven’t packaged this version. but we will!

    https://github.com/teaxyz/pantry.core/issues/new
```

PR #356, opened 2023-02-03
fix: using memory proportional to download size

### LINKED ISSUE #338: tea consuming RAM in proportion to download size
When I run `tea +rust-lang.org` on a fresh Codespace, the command seems to begin installation, but then gets killed. This may be a problem with codespaces, but if it is, I have never seen it before.
image
```
$ sh <(curl tea.xyz) +rust-lang.org
100 13306  100 13306    0     0   316k      0 --:--:-- --:--:-- --:--:--  316k
llvm.org=14.0.6 downloading 31%
Killed
```

Regular rust installation works fine in codespaces.
It seems that llvm.org installation gets to 31% or so and then is `Killed`

lino-levan:
Sorry about waiting so long to do this easy PR. I have been relatively busy this week and technically I am still busy but I felt bad about promising to do this and not following through.

Theoretically closes #338

### DIFF
--- a/import-map.json
+++ b/import-map.json
@@ -6,7 +6,7 @@
     "hooks/": "./src/hooks/",
     "prefab": "./src/prefab/index.ts",
     "prefab/": "./src/prefab/",
-    "deno/": "https://deno.land/std@0.173.0/",
+    "deno/": "https://deno.land/std@0.176.0/",
     "semver": "./src/utils/semver.ts",
     "utils": "./src/utils/index.ts",
     "utils/": "./src/utils/",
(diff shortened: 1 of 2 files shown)


lino-levan:
It is quite unlucky but I believe this is the only non-ffi solution to this problem

mxcl (approved):

mxcl:
Thank you so much @lino-levan! It would have taken us much longer to debug this.

mxcl:
Sorry I didn't merge this sooner, it didn’t hit my notifications radar for some reason. Pushing out a point release now.

PR #360, opened 2023-02-04
implement tea-magic in rust

jhheider:
resolves https://github.com/teaxyz/cli/issues/341, maybe

```sh
$ hyperfine --warmup=3 "tea --magic" 
Benchmark 1: tea --magic
  Time (mean ± σ):      47.5 ms ±   1.3 ms    [User: 43.3 ms, System: 3.4 ms]
  Range (min … max):    46.2 ms …  53.8 ms    60 runs
 
$ hyperfine --warmup=3 "./tea-magic zsh"
Benchmark 1: ./tea-magic zsh
  Time (mean ± σ):       0.6 ms ±   0.5 ms    [User: 0.3 ms, System: 0.2 ms]
  Range (min … max):     0.2 ms …   9.4 ms    1084 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```

slight performance gains. would need support in `teaxyz/setup` and `teaxyz/pantry.core`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,7 +431,7 @@ shell’s configuration file.
 Our magic is not automatically added to scripts, but you can manually add it:
 
 ```sh
-source <(tea --magic=bash)
+source <(tea-magic bash)
                    # ^^ you have to specify which shell tho
 ```
 
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -11,8 +11,10 @@
     "run": "deno run --unstable --allow-all src/app.ts",
     // compiles to ./tea
     "compile": "deno compile --allow-read --allow-write --allow-net --allow-run --allow-env --unstable --output $INIT_CWD/tea src/app.ts",
+    // compiles to ./tea-magic
+    "compile-magic": "rustc -C opt_level=3 -o $INIT_CWD/tea-magic src/tea-magic/main.rs",
     // installs this source checkout as a tea stowed package
-    "install": "deno compile --unstable -Ao $TEA_PREFIX/tea.xyz/v$VERSION/bin/tea src/app.ts && scripts/repair.ts tea.xyz"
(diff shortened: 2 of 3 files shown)


vegerot (approved):

mxcl:
So we can package this as `tea.xyz/magic` and version it and all that. Then the installer needs updating. Then we have to make it so the installer changes the current shell rc line to the new one. And we have to make the installer update logic do that too.

Overall it's a lot of complexity to split tea into 2 pieces. And it means our “we're a standalone binary download” messaging is now a little fuzzier.

## Counters:

* `deno` claim binary cold start performance is going to improve.
* We would still like to rewrite in rust once “stable” (1.0 is planned for March)
* The main source of slowdown is our `cd` hook, not `--magic` (which only happens for new terminal instantiations)

## Conclusions

With such near timelines for the 2 counters,  the complexity of splitting ourselves into 2 binaries at this time is quite high and risky. I'd rather not.
\
## Alternatives

`--magic` is very cacheable. A possible easy and safe solution is to do cache based on the ctime of the resolved symlink `~/.tea/tea.xyz/v*` thus when we install a new tea version it will recompute the cache.

jhheider:
> Thanks for the work all the same 😔

Of course. Totally worth it to have some actual test numbers.

mxcl:
I know you are always looking for an excuse to fire up rust anyway :D

jhheider:
I mean, that can't be why I've written two small CLI tools in it recently. Makes no sense. ;)

PR #387, opened 2023-02-17
0.25

mxcl:


src/app.magic.ts
```diff
@@ -21,8 +21,10 @@ export default function(self: Path, shell?: string) {
         chpwd_functions=( _tea_chpwd_hook \${chpwd_functions[@]} )
       fi
 
-      # if the user put tea in eg. /usr/local/bin then don’t pollute their PATH
-      # we check for \`tea --prefix\` due to \`gitea\` being \`tea\` when installed with \`brew\`
+      if [[ "$PATH" != *"$HOME/.local/bin"* ]]; then
```
jhheider, line 24: There's an extremely minor edge-case here, where this will skip if they have `$HOME/.local/bin/foo` in their path. A better test (with more dependencies) is `echo $PATH | tr ':' '\n' | grep -E "^$HOME/.local.bin$"`

jhheider: Or `split` as you do below.

mxcl: You’re right. I think this will have to do for now though as really trying to only use builtins for speed and portability.

jhheider:
Well, that wasn't intended. I'll back all this up.

PR #411, opened 2023-02-27
don't constrain cyclic deps to build version

jhheider:
requires #406 to pass checks

fixes https://github.com/teaxyz/cli/issues/412

Basically if a dep requests the target (`go.dev` requires `go.dev: '*'`), this will prevent locked loop (build `go.dev=1.20.1` trying to install `go.dev=1.20.1`)

### DIFF
--- a/src/prefab/hydrate.ts
+++ b/src/prefab/hydrate.ts
@@ -61,6 +61,10 @@ export default async function hydrate(
 
       for (const dep of await get_deps(node.pkg, initial_set.has(node.project))) {
         if (children.has(dep.project)) {
+          const self = graph[dep.project]
+          if (self.pkg.constraint === target.pkg.constraint) {
+            self.pkg.constraint = dep.constraint
+          }
           if (!bootstrap.has(dep.project)) {
             console.warn(`${dep.project} must be bootstrapped to build ${node.project}`)
 


src/prefab/hydrate.ts
```diff
@@ -61,6 +61,10 @@ export default async function hydrate(
 
       for (const dep of await get_deps(node.pkg, initial_set.has(node.project))) {
         if (children.has(dep.project)) {
+          const self = graph[dep.project]
+          if (self.pkg.constraint === target.pkg.constraint) {
```
mxcl, line 65: seems like you need to implement `Range.eq`, surely `===` will only work here if the object itself is the same object?

mxcl:
k, so changing this code is dangerous. We better get it right.

Can you highlight what part of the code is failing per your failure spec?

src/prefab/hydrate.ts
jhheider: This seems to do exactly what I want, in testing. Replaces the build pkg with the dep so we don't require the version we want to build.

jhheider:
> Can you highlight what part of the code is failing per your failure spec?

No patch:
```sh
$ ./libexec/deps.ts go.dev~1.20 --build
go.dev must be bootstrapped to build go.dev
curl.se/ca-certs
gnu.org/m4^1
tea.xyz/gx/cc^0.1
openssl.org^1
go.dev~1.20
```

Patched:
```sh
$ ./libexec/deps.ts go.dev~1.20 --build
go.dev must be bootstrapped to build go.dev
curl.se/ca-certs
gnu.org/m4^1
tea.xyz/gx/cc^0.1
openssl.org^1
go.dev
```

`go.dev` requires `go.dev: '*'` to build. *But* it shouldn't require the explicit version being built (that's madness). Same for `llvm.org` on linux, iirc. Likely the others that are self-requiring.

mxcl:
I think the actual solution for the this is to hydrate the deps of go rather than go itself.

So this is a regression in brewkit.

PR #418, opened 2023-02-28
Issue#324 - add test coverage

### LINKED ISSUE #324: enable code coverage
deno supports `--coverage`, then upload to whatever isn’t as terrible as codecov.io

ABevier:
Adding test coverage with coveralls.io.  Coverage is only counted for the ubuntu tests.  

fixes #324

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,10 +24,10 @@ jobs:
       - run: deno task test --coverage=cov_profile
       - run: deno coverage cov_profile --lcov --output=cov_profile.lcov
       - name: coveralls
-          uses: coverallsapp/github-action@master
-          with:
-            github-token: ${{ secrets.GITHUB_TOKEN }}
-            path-to-lcov: cov_profile.lcov
+        uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          path-to-lcov: cov_profile.lcov
 
   # run tests for non-ubuntu OS's
   tests:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,22 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # run tests + coverage for unbuntu
+  coverage:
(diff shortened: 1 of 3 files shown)


jhheider:
A different way to do it would be to add the coverage to the tests and use an if: key to only run it for one arm of the matrix.

ABevier:
> A different way to do it would be to add the coverage to the tests and use an if: key to only run it for one arm of the matrix.

I considered that, but I decided against it because it seemed like it would make the job a little busy, but I can change it if you'd like.

jhheider:
The main benefit I see is that it lets us keep the matrix working, instead of having one test elsewhere. Alternatively, running the test twice isn't the end of the world, either. I don't think there's any wrong answer here, just a matter of what people like.

mxcl:
well, we should probably upload for both Mac and Linux anyway. We have platform specific pathways that we'll want to flag for coverage.

jhheider:
Coveralls also produces [badges](https://github.com/jhheider/semverator/blob/618d6c95b5b6c95d1061c6175ebe1f4b0475929f/README.md?plain=1#L4), but seeing our 3% number might feel a little bad ;)

jhheider:
> well, we should probably upload for both Mac and Linux anyway. We have platform specific pathways that we'll want to flag for coverage.

this is an interesting issue. i think coveralls is just last-upload-is-the-truth, so stomping one arch with the other isn't perfect. it'd be nice to mock platform in the tests, but i bet it's not feasible.

ABevier:
> > well, we should probably upload for both Mac and Linux anyway. We have platform specific pathways that we'll want to flag for coverage.
> 
> this is an interesting issue. i think coveralls is just last-upload-is-the-truth, so stomping one arch with the other isn't perfect. it'd be nice to mock platform in the tests, but i bet it's not feasible.

They support sharding tests and then merging the results, I'm trying that out, but i'm not sure what's going to happen when the results stomp each other.

ABevier:
Coveralls tracks the coverage of each job and appears to combine the results in sane way:

Screenshot 2023-02-28 at 6 30 38 PM

PR #428, opened 2023-03-07
Add user-agent header

ABevier:
I created a useFetch hook that wraps the Deno `fetch` function and automatically adds a `User-Agent` header to the request.

The header looks like `User-Agent: tea.cli/0.24.9`

### DIFF
--- a/src/hooks/useFetch.ts
+++ b/src/hooks/useFetch.ts
@@ -1,0 +1,8 @@
+import { useVersion } from "hooks";
+
+// useFetch wraps the native Deno fetch api and inserts a User-Agent header
+export default function useFetch(input: string | URL | Request, init?: RequestInit | undefined): Promise<Response> {
+    const requestInit = init ?? {} as RequestInit
+    requestInit.headers = { ...requestInit.headers, "User-Agent": `tea-cli/${useVersion()}` }
+    return fetch(input,  requestInit)
+}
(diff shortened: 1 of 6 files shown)


mxcl (approved):

src/hooks/useFetch.ts
```diff
@@ -0,0 +1,8 @@
+import { useVersion } from "hooks";
+
+// useFetch wraps the native Deno fetch api and inserts a User-Agent header
+export default function useFetch(input: string | URL | Request, init?: RequestInit | undefined): Promise<Response> {
+    const requestInit = init ?? {} as RequestInit
+    requestInit.headers = { ...requestInit.headers, "User-Agent": `tea-cli/${useVersion()}` }
```
mxcl, line 6: Can we go with `tea/cli/version` or is that violating expectations for parsers of the user-agent?

Maybe `tea (cli)` is more typical.

Either way I'm not fond of the dash here.

ABevier: I think `tea(cli)/version` is probably fine if a bit atypical.  

You can see a bunch of examples here:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

mxcl: yeah you're right. Parenthesis are used for platform info not variants.

yeah, back to `tea-cli/version` I agree with your original assessment lol

mxcl: Or perhaps `tea.cli/version` which we've used elsewhere sometimes.

ABevier: tea.cli it is

PR #438, opened 2023-03-13
increase test coverage

ABevier:
This is the first of a few PR's that will start to make the codebase more testable.  Future PR's will include:
- Moving global state to a useConfig hook
- Moving most Deno interactions to hooks that can be mocked more easily
- Moving all printing logic behind a usePrint hook


import-map.test.json
```diff
@@ -0,0 +1,22 @@
...
+  "imports": {
+    "hooks/usePrefix.ts": "./tests/mocks/mockUsePrefix.ts",
+    "hooks/usePrint.ts": "./tests/mocks/mockUsePrint.ts",
+    "path": "./src/vendor/Path.ts",
+    "types": "./src/types.ts",
+    "hooks": "./src/hooks/index.ts",
+    "hooks/": "./src/hooks/",
+    "prefab": "./src/prefab/index.ts",
+    "prefab/": "./src/prefab/",
+    "deno/": "https://deno.land/std@0.176.0/",
```
jhheider, line 11: I wish there was a way to extend/overlay import maps, so we didn't have to maintain a dep in two places. I particular, I've run into things where I need a remote import map + a local map, or two remote maps. This would be very nice to have in deno.
```json
"extends": "./src/import-map.json",
"hooks/": "./tests/mocks/hooks/",
```

run_tests.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+rm -rf cov_profile
+deno task test --coverage=cov_profile 
+deno coverage cov_profile --lcov --exclude=tests/ > cov_profile/coverageFile.lcov
+genhtml -o cov_profile cov_profile/coverageFile.lcov
+open cov_profile/index.html
```
jhheider, line 7: This should probably be a `task` in `deno.jsonc` for simplicity's sake. Also, the shell redirect can be replaced with `--output=cov_profile/coverageFile.lcov` if I'm reading the `--help` correctly.

src/app.dump.ts
```diff
@@ -1,5 +1,6 @@
 import { EnvKeys } from "hooks/useShellEnv.ts"
-import { flatmap, print } from "utils"
+import { usePrint } from "hooks";
```
jhheider, line 2: omit `;` for consistency (I know the auto-complete likes to add them).

src/app.dump.ts
```diff
@@ -45,35 +46,35 @@ export default async function dump({ env }: Parameters) {
     if (Deno.env.get("TEA_REWIND")?.includes("TEA_PREFIX")) {
       // we do this manually because we also set it for stuff executed through tea
       // which means when developing tea this otherwise doesn’t get unset lol
-      await print(unsetEnv("TEA_PREFIX"))
+      await usePrint(unsetEnv("TEA_PREFIX"))
```
jhheider, line 49: From an API perspective, I feel like `usePrint()` should return an object containing a `print()` function. I believe it's generally the case that hooks are of a form of `function useSomething(optional_initial_state): object_of_functions_and_values`. At least in `cli`'s codebase.

jhheider: Ironically, such minor complexity would reduce your diff size, though not very meaningfully.

tests/mocks/mockUsePrefix.ts
```diff
@@ -0,0 +1,13 @@
+import { default as realUsePrefix, Prefix } from "../../src/hooks/usePrefix.ts"
+import Path from "path";
+
+let mockPrefixPath: Path | null = null;
```
jhheider, line 4: I think we'd generally use `let mockPrefixPath: Path | undefined`

tests/mocks/mockUsePrefix.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const setTeaPrefix = (prefix: string) => mockPrefixPath = new Path(prefix);
```
jhheider, line 6: `Path.constructor()` takes `string | Path`, so you can make this slightly more general (if desired) for free.

jhheider:
Looks like a really good basis.

import-map.test.json
ABevier: When I finish the refactor I think it probably won't need this at all anyways.  Deno docs recommend using an `_internal` function table for doing mocks and once we've consolidated our a lot of the deno calls into the hooks mocking will be much easier.

src/hooks/usePantry.ts
```diff
@@ -227,13 +228,16 @@ function tokenizePackage(pkg: Package) {
...
 export function pantry_paths(): Path[] {
   const rv: Path[] = []
+  const { TEA_PANTRY_PATH } = useEnv()
+
+  const prefix = getPantryPrefix()
   if (prefix.isDirectory()) rv.push(prefix)
-  const env = Deno.env.get("TEA_PANTRY_PATH")
-  if (env) for (const path of env.split(":").reverse()) {
+
+  if (TEA_PANTRY_PATH) for (const path of TEA_PANTRY_PATH.split(":").reverse()) {
```
mxcl, line 240: you are allowed to think this sort of thing is why we need a linter 😂

src/hooks/usePrefix.ts
```diff
@@ -1,31 +1,7 @@
...
-    const shelf = new Path(Deno.execPath())
-      .readlink() // resolves the leaf symlink (if any)
-      .parent()
-      .parent()
-      .parent()
-
-    switch (shelf.basename()) {
-    case 'tea.xyz':
-    case 'deno.land':
-      return shelf.parent()
```
mxcl, line 19: I wonder if we still need this bit. We migrated to “you must define TEA_PREFIX if you aren't in ~/.tea” kinda thing

mxcl (approved):
Didn't review the new tests.

Excellent work. A truly inspiring refactor remarkable in its conservativeness and scope.

src/app.dump.ts
mxcl: I decided to overlook these as we've all done it.

scripts/run_coverage.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e
+
+rm -rf cov_profile
+deno task test --fail-fast --coverage=cov_profile 
+deno coverage cov_profile --lcov --exclude=tests/ --output=cov_profile/coverage_file.lcov
+genhtml -o cov_profile cov_profile/coverage_file.lcov
```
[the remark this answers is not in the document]
ABevier: genhtml comes from lcov.  I looked at packaging it but it's written in Perl which scared me away

PR #441, opened 2023-03-14
Test coverage 2

ABevier:
This PR refactors the application so that all Global state flows through the `useConfig` hook.  Environment variables are collected at startup and provided via the `useEnv` hook.  This allows the application to much more easily change configuration during tests.

`useFlags` has been removed in favor of `useConfig` and `useEnv`


src/app.main.ts
```diff
@@ -11,18 +9,19 @@ import Path from "path"
...
   if (args.cd) {
     const chdir = args.cd
     console.verbose({ chdir })
     Deno.chdir(chdir.string)
   }
 
-  if (!Deno.isatty(Deno.stdout.rid) || Deno.env.get('CI')) {
-    logger.set_global_prefix('tea:')
-  }
-
```
jhheider, line 25: Is this reproduced somewhere else (didn't see in the diff).

src/hooks/useConfig.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  verbosity: Verbosity
+  dryrun: boolean
+  keepGoing: boolean
+
+  verbose: boolean
+  debug: boolean
+  silent: boolean
+
+  env: Env
```
jhheider, line 43: I wonder if this (or `Env`) needs a `raw: string[]` value for future sanity. If we don't need it now, then maybe we're fine.

src/hooks/useLogger.ts
```diff
@@ -1,13 +1,7 @@
...
-import useFlags from "./useFlags.ts"
-
-let global_prefix: string | undefined
-export function set_global_prefix(prefix: string) {
-  // FIX ME!!!
-  //if (global_prefix !== undefined) throw new Error()
-  global_prefix = prefix.trim()
-}
+import useConfig, { useEnv } from "hooks/useConfig.ts";
+import { Verbosity } from "../types.ts";
```
jhheider, line 4: `from "types"`

src/hooks/useLogger.ts
```diff
@@ -33,30 +27,32 @@ function ln(s: string, prefix_length: number) {
...
-    if ((Deno.env.get("NO_COLOR") ?? '0') != '0') {
+    if ((NO_COLOR ?? '0') != '0') {
       return false
     }
-    if (Deno.env.get("CLICOLOR") == '0' || Deno.env.get("CLICOLOR_FORCE") == '0') {
+    if (CLICOLOR == '0' || CLICOLOR_FORCE == '0') {
       return false
     }
-    if (Deno.env.get("CI")) {
+    if (CI) {
```
jhheider, line 55: `isCI`, since boolean context?

jhheider (approved):
This looks like a clean refactor. Basically moves code/alters APIs.

src/app.dump.ts
```diff
@@ -14,7 +14,8 @@ interface Parameters {
 }
 
 export default async function dump({ env }: Parameters) {
-  const shell = flatmap(Deno.env.get("SHELL"), basename)
+  const { SHELL, TEA_REWIND } = useEnv();
+  const shell = flatmap(SHELL, basename)
```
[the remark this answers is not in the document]
ABevier: So my thought was to do all caps because they were env vars (and it was easier to change the spots where we used `Deno.env.get`) but yeah maybe that wasn't a great idea.  I could go either way.

src/app.main.ts
ABevier: It's handled in `useLogger` which pulls from `useConfig` to avoid setting additional global state.

https://github.com/teaxyz/cli/pull/441/files#diff-51338d7c9f90a73701d0b7aac4de25dc47ed8f7d1407ee7f276e17840ea45302R30-R31

src/hooks/useConfig.ts
ABevier: raw would need to return `{key: string, value: string}[]` right?

src/app.dump.ts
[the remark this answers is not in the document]
jhheider: I as well. The undecorated ALL-CAPS vars look weird, but if you know why they're all-caps, it's communicative. @mxcl you have a horse in this race?

src/hooks/useConfig.ts
jhheider: hm. yes. or possibly `Record<string, string>[]`.

PR #451, opened 2023-03-17
fix magic for bash

### LINKED ISSUE #450: panic:Broken pipe (os error 32)
# tea does not run after installation

## Description

I tried to use tea on MacOS.
I got an error by removing `--silent` flag on tea.

## Environment

MacOS 13.2.1
tea 0.25.1

## Reproduction

`sh <(curl https://tea.xyz) && source <("$HOME/.tea/tea.xyz/v*/bin/tea" --magic=bash)`

## Result

```
$ sh <(curl https://tea.xyz) && source <("$HOME/.tea/tea.xyz/v*/bin/tea" --magic=bash)
100 13557  100 13557    0     0   138k      0 --:--:-- --:--:-- --:--:--  142k

  hi 👋 let’s set up tea

  • we’ll put it here:  /Users/jwel/.tea
  • everything tea installs goes there
  • (we won’t touch anything else)

  │ docs https://github.com/teaxyz/cli#getting-started

grep: /Users/jwel/.bashrc: No such file or directory

  may we interest you in some magic?

  tea’s shell magic is optional but it’s the way it’s meant to be used.

  │ docs https://github.com/teaxyz/cli#magic

  Added:

   test -d "$HOME/.tea" && source <("$HOME/.tea/tea.xyz/v*/bin/tea" --magic=bash --silent)


  you’re all set!

  try it out:

   exec bash -i   # or open a new tab
   wget -qO- tea.xyz/white-paper | glow -

panic: spilt tea. we’re sorry and we’ll fix it… but you have to report the bug!

    https://github.com/teaxyz/cli/issues/new?title=panic%3ABroken+pipe+(os+error+32)

----------------------------------------------------->> attachment begin
BrokenPipe: Broken pipe (os error 32)
    at async file:///private/var/folders/8b/n_gmhmk57ks85swq3mrdbrwc0000gp/T/7a834af1/src/app.ts:111:13
<<----------------------------------------------------- attachment end
```

ABevier:
Magic currently fails on bash 3.2 (the default version of bash that ships with macOS).
This is due to a [bug in bash 3.2 related to sourcing from a process substitution.](https://lists.gnu.org/archive/html/bug-bash/2006-01/msg00018.html)

This is a known workaround that will work for bash 3.x and newer versions.

Note that although this fixes the sourcing issue, there are other issues with magic when running bash 3.2 (`command_not_found_handler` doesn't work)
 
Closes #450

### DIFF
--- a/src/app.magic.ts
+++ b/src/app.magic.ts
@@ -99,7 +99,7 @@ export default function(self: Path, shell?: string) {
     return undent`
       cd() {
         builtin cd "$@" || return
-        source <("${d}"/tea +tea.xyz/magic -Esk --chaste env)
+        source /dev/stdin <<<"$("${d}"/tea +tea.xyz/magic -Esk --chaste env)"
       }
 
       if [[ "$PATH" != *"$HOME/.local/bin"* ]]; then


mxcl:
One issue is that the test for validating magic is ok *was* passing before this. Will merge, but we need to fix that.

ABevier:
> One issue is that the test for validating magic is ok _was_ passing before this. Will merge, but we need to fix that.

@mxcl  this fix only affects bash version 3.2.  The test uses bash 5.X.  We could add a test that forces the check with 3.2 but that won't work because tea has other issues running on bash 3.2 that I previously mentioned.

Ideally, if you find a way to add support for bash 3.2 a proper test would be added then.

PR #452, opened 2023-03-18
Integration tests should catch fish syntax errors

ABevier:
Fish doesn't have an equivalent of bash's `set -e` so even though it fails to source tea's magic, the script continues to run. 

There isn't a great way to detect an error when sourcing a fish script because the exit code of `source` is only for the last command run.  I added a check in the test for anything going to stderr that wasn't from tea to help catch syntax errors.

Unfortunately as I needed to check both stderr and stdout I also refactored the integration test suite so that the exit code, stdout, and stderr are always available.  This change is actually the bulk of the PR.

### DIFF
--- a/tests/integration/magic.test.ts
+++ b/tests/integration/magic.test.ts
@@ -100,3 +100,24 @@ it(suite, "tea --magic in a script. fish", async function() {
 
   await this.run({ args: [script.string] })
 })
+
+
+it(suite, "tea verify --magic is parsed correctly by fish", async function() {
+  const script = this.sandbox.join("magic-fish").write({ text: undent`
+    #!/usr/bin/fish
+
+    tea --magic=fish | source
+
+    `})
+
+  // fish doesn't have an equivalent of bash's "set -e" to exit the script if an error occurs
+  // for more information go here: https://github.com/fish-shell/fish-shell/issues/510
+  // we will assume if stderr contains anything that isn't prefixed with tea that then 
+  // an error occurred
+  const stderr = await this.run({ args: [script.string] }).stderr()
+
+  const errorLines = stderr.split("\n")
+    .compact(x => strip_ansi_escapes(x).trim())
+    .filter(x => !x.startsWith("tea:"))
+  assertEquals(errorLines, [])
+})


mxcl:
Doesn't this mean that stderr/stdout are always suppressed? If so, this will make debugging failing tests where we don’t need stdout and stderr awful.

ABevier:
> Doesn't this mean that stderr/stdout are always suppressed? If so, this will make debugging failing tests where we don’t need stdout and stderr awful.

No it's the opposite, stdout default to on, but can be disabled if necessary.  The original implementation defaulted to suppressing stdout and stderr unless they were specifically requested by awaiting the promise.  However, the way the promise was implemented wasn't typesafe and only allowed for stdout OR stderr OR the exit code to be returned.  

I added the option to disable it only because of the original implementation, otherwise I think that the test harness should just always collect everything.

ABevier:
> Doesn't this mean that stderr/stdout are always suppressed? If so, this will make debugging failing tests where we don’t need stdout and stderr awful.

Okay, after rereading this comment again with fresh eyes I misunderstood and my previous answer was wrong.  What if I just console.log and console.error stdout and stderr if they are piped?

ABevier:
Once either stdout or stderr are piped to Deno it doesn't appear that there is any reliable way to maintain ordering.  I can't easily supplement what already existed because it only allows piping a single result and is subject to a race depending on how the promise is awaited.  There is a possibility to require a `pipeStdout` and `pipeStderr` param in order to safely signal that output should be piped, but that's a poor developer experience imho.

mxcl:
Instead of rewriting all this test infra I have done this locally as the test for fish magic:

```fish
    test $(basename $(ps -hp $fish_pid | awk "{print \\$$N}" | tail -n1)) = fish

    tea --magic=fish | source

    export NODE_DISABLE_COLORS=1
    export CLICOLOR_FORCE=0
    export VERBOSE=-1  # no tea output FIXME doesn’t seem to work…?
    node^18 --eval "console.log('xyz.tea.hi')" > node.out

    if test "$(cat node.out)" != xyz.tea.hi
      exit 3
    end

    cd .
    node --version
```

Which will work I assume. However it exposes that our fish patch to stop the command not found handler from outputting 127 rather than the command we run’s exit code is no longer working. I'm looking in to that in another ticket.

Anyway, anything obvious I am missing in the above? It uses the fish script we run itself to exit if there's bad behavior thus avoiding having to do hacks or rewrite any infra.

mxcl:
magic should be a pantry thing anyway, then we could validate this there as part of pkg tests

ABevier:
So while your updated script will work, I'm not sure it's the best method to handle this.  As fish doesn't have an equivalent to `set -e` you will need to ensure that every line of the magic applied successfully with the test.  If part of the magic fails to parse, the other parts will still source correctly, so this test will need to be completely comprehensive.

Checking stderr for failures to source seems more robust to me in the long run and doesn't solve the root problems with the integration test suite:  you can only request one of (exit code | stdout | stderr)

mxcl:
> Checking stderr for failures to source seems more robust to me in the long run

stderr from arbitrary packages that we have no control over doesn’t really fit my definition of robust 😏.

However now I revisit the original bug I see that indeed we don’t check all the magic.

Why not add another fish test that simple calls the magic and then checks the stderr?

mxcl:
See #465 for changes. I see now thanks to you making me revist this that actually it doesn’t really validate anything new.

ABevier:
> Why not add another fish test that simple calls the magic and then checks the stderr?

I suppose in the spirit of compromise, this is the correct answer.

mxcl (approved):
mxcl.APPROVE()

PR #457, opened 2023-03-20
beef up test coverage

ABevier:
Adding more coverage with the functional test framework.


src/utils/pkg.ts
```diff
@@ -21,10 +21,10 @@ export function parse(input: string): PackageRequirement {
       const parts = v.split(".")
       const n = parts.length
       switch (n) {
-      case 0:
+      case 1:
```
mxcl, line 24: what's this all about then?

src/utils/safe-utils.ts
```diff
@@ -7,11 +7,3 @@ export function chuzzle(input: number) {
 export function panic(message?: string): never {
   throw new Error(message)
 }
-
-export function flatmap<S, T>(t: T | undefined | null, body: (t: T) => S | undefined, opts?: {rescue?: boolean}): NonNullable<S> | undefined {
```
mxcl, line 11: my beloved flatmap!?!?!

src/vendor/Path.ts
```diff
@@ -276,7 +276,7 @@ export default class Path {
 
   mkdir(): Path {
     if (!this.isDirectory()) {
-      Deno.mkdirSync(this.string)
+      Deno.mkdirSync(this.string, { recursive: true })
```
mxcl, line 279: but we have `mkpath()` for that.

while I agree I don't particularly see many reasons to not always `mkdir -p` I feel we should meet expectations and based on history that is mkdir doesn't recurse.

mxcl (changes requested):
not many changes (from me). Also didn't review the tests carefully. Anything I should look at carefully?

src/vendor/Path.ts
jhheider: I hate that feature of `mkdir`, but it feels safer for it to be explicit.

src/utils/safe-utils.ts
mxcl: I guess we no longer use it. So indeed removal is the right thing.

For coverage 🫡

src/vendor/Path.ts
mxcl: like really we should have `mkdir` and have it have a `p` option. Using the string `'p'` because I have oddly become attached to the idea of this library mirroring the cli commands completely.

jhheider: [you're not the only one](https://www.npmjs.com/package/mkdirp)

src/utils/safe-utils.ts
ABevier: it's in the codebase twice actually, I removed the unused version.

src/utils/pkg.ts
ABevier: split never returns 0 parts.  "6".split(".") returns ["6"]

src/utils/safe-utils.ts
jhheider: He loves it so much.

ABevier:
>not many changes. Also didn't review the tests carefully. Anything I should look at carefully?

Devenv tests are probably the most important `tests/functional/devenv.test.ts` and take another look at the SemVer parsing for `@` I think I fixed an off by 1 error there, but now I'm second guessing myself.

src/utils/safe-utils.ts
mxcl: lol. incredible

src/utils/pkg.ts
mxcl: ofc. how did the tests pass?!?!?

Are we protected against `"".split('.')` that always catches me in js

mxcl:
beautiful work.

Honestly remarkable you managed to decode what all that shit was doing so quickly and effectively.

PR #459, opened 2023-03-21
Rename pantry.extra and pantry.core to pantry

magnusviri:
I noticed some instances of pantry.extra and pantry.core. If I understand correctly, these should all be just "pantry" now.


mxcl:
k 0.26 merged and indeed this PR is now empty. Closing. But thank you.

PR #469, opened 2023-03-24
coverage script should use tea

ABevier:
Now that `lcov` is packaged this script can be updated to use tea

### DIFF
--- a/scripts/run_coverage.sh
+++ b/scripts/run_coverage.sh
@@ -1,4 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env tea
+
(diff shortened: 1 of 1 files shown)


scripts/run_coverage.sh
```diff
@@ -1,4 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env -S tea
```
jhheider, line 1: Shouldn't need `-S` (--split) for a single value (`tea`). Won't hurt though.

jhheider (approved):
Fantastic.

PR #472, opened 2023-03-24
[ci] WSL2

ABevier:


mxcl:
https://github.com/marketplace/actions/setup-wsl WSL2 not explicitly supported.

So we're going to need a self-hosted runner with Windows 10/11 with WSL2 preinstalled. @jhheider thoughts?

jhheider:
Cheap windows box? Or an EC2 or Azure VPS. Assuming we add to pantry as well, dedicated hardware will give better return if we generate enough minutes.

mxcl:
I was thinking of hosting it virtually on our existing runners, but I'm not fussed. If hardware is required then get waters to order it.

jhheider:
we could run it virtually on the x86-64 box, but i worry about continuing to subdivide those resources. let me try and spin up a virtual runner and see what kind of timing we're looking at.

mxcl:
WSL2 is identical to whatever linux you install inside it. So… this isn’t necessary IMO.

PR #478, opened 2023-03-26
Detect, preserve and render calver

mxcl:
@jhheider worried this may break kettle side stuff etc. what you think?

Might well be a dumb thing to bother with.

### DIFF
--- a/src/utils/semver.ts
+++ b/src/utils/semver.ts
@@ -43,7 +43,34 @@ export default class SemVer {
         } else {
           throw new Error(`invalid version: ${input}`)
         }
-      })
+      });
+      this.pretty = (() => {
+        // detect calver (eg. 2023.03.01)
+        // we also accept 0.0.0.2023.03.01 because this is how we version pre-release software
+        // we version pre-release software this was so when they finally release a version
+        // it will be versioned higher than the pre-releases
+
+        const cc = [...this.components]
+        if (cc[0] == 0) cc.shift()
+        if (cc.length != 3) return
+        if (this.components[0] < 1990) return  // anything before 1990 can go fish
+        if (this.components[1] < 1) return
+        if (this.components[1] > 12) return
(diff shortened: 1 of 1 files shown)


jhheider:
Worth a try. I think you need < 1 not < 0 for the second and third terms.

mxcl:
The problem here is downloading stuff we already calver’d but without left-padded zeros.

Everything else is fine. So if I wanted this badly enough I'd need to rename the stuff S3 or put some redirects in.

Then it's whether or not the kettle uses this code correctly?

jhheider:
> The problem here is downloading stuff we already calver’d but without left-padded zeros.
> 
> Everything else is fine. So if I wanted this badly enough I'd need to rename the stuff S3 or put some redirects in.
> 
> Then it's whether or not the kettle uses this code correctly?

kettle uses brewkit:usePantry to feed versions to pantry:new-version. the latter is more likely to cry than the former (but if it works in brewkit, it should be fine).

src/hooks/useOffLicense.ts
```diff
@@ -14,7 +14,7 @@ function key(stowage: Stowage) {
     const { platform, arch } = stowage.host ?? host()
     rv = rv.join(`${platform}/${arch}`)
   }
-  let fn = `v${stowage.pkg.version}`
+  let fn = `v${stowage.pkg.version.raw ?? stowage.pkg.version}`
```
mxcl, line 17: k so… using version.raw means we use what was recording in versions.txt.

However I fear that `versions.txt` regenerates every time and uses `SemVer.toString()` rather than the raw version from the filename.

So this will break in prod when packages update.

I would like this because rendering calver properly is something I enjoy but I am unwilling to cause breakage.

jhheider: Praps `SemVer.toString()` needs calver preservation?

mxcl:
not worth it

PR #482, opened 2023-03-27
test coverage > 80% 🤞 

ABevier:
Removed some unused functions in utils and added tests for `Path` to try to get to 80%

### DIFF
--- a/src/vendor/Path.ts
+++ b/src/vendor/Path.ts
@@ -340,9 +340,21 @@ export default class Path {
     return Deno.readTextFile(this.string)
   }
 
-  readLines(): AsyncIterableIterator<string> {
+  async *readLines(): AsyncIterableIterator<string> {
     const fd = Deno.openSync(this.string)
-    return readLines(fd)
+    const lines = readLines(fd)
+    try {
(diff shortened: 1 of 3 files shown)


src/vendor/Path.ts
```diff
@@ -340,9 +340,21 @@ export default class Path {
...
+    const lines = readLines(fd)
+    try {
+      while (true) {
+        const res = await lines.next()
+        if (res.done) {
+          break
+        }
+        yield res.value
+      }
+    }
```
jhheider, line 354: Does this need to re-throw if there's an error? And would this work more simply?
```typescript
for await(const line of lines) {
  yield line.value
}
fd.close()
```
or will it actually throw?

jhheider (approved):

src/vendor/Path.ts
ABevier: yeah that's much better, I even did it that way in the test code.  And the try/finally will allow the error to continue up the stack while correctly freeing the file resource, so catch and rethrow is no necessary.

mxcl (approved):

PR #494, opened 2023-03-30
TEA_PKGS is not read correctly

ABevier:
Dev envs were not correctly restricting tea packages to the correct versions because of this typo.

PR #496, opened 2023-03-30
Add tests for collectEnv

ABevier:
This PR adds tests to prevent the regression of the bug fixed by #494

### DIFF
--- a/tests/functional/args.test.ts
+++ b/tests/functional/args.test.ts
@@ -1,7 +1,7 @@
 import { assertEquals } from "deno/testing/asserts.ts"
 import { wut } from "../../src/app.main.ts"
 import { parseArgs } from "../../src/args.ts"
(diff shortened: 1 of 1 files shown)


tests/functional/args.test.ts
```diff
@@ -101,3 +102,31 @@ Deno.test("wut args", async test => {
     assertEquals(runTest(["+tea.xyz/magic", "env"]), "dump")
   })
 })
+
+Deno.test("reads env", () => {
+  const keys = ["CI", "CLICOLOR", "CLICOLOR_FORCE", "DEBUG", "GITHUB_ACTIONS", "GITHUB_TOKEN",
```
mxcl, line 107: probs should add a commend in the file where we declare these. Potentially can use typescript to generate the array

mxcl (approved):

PR #501, opened 2023-04-05
dark magic v0

jhheider:
picked the easiest one first
- if no other resolution fails, query `registry.npmjs.org`
- if that finds a package, return `["npx", "-q"]` as the shebang
- hookify it all to make extension simple
- use `Promise.all` to parallelize the search


mxcl (approved):

jhheider:
Tests pass, test added for new functionality, and @mxcl already approved it, so it's YOLO time.

src/hooks/useExec.ts
```diff
@@ -287,6 +287,18 @@ export async function which(arg0: string | undefined) {
     return found
   }
 
+  // dark-magic will find a _lot_ of things, even when we might not want to.
+  // https://www.npmjs.com/package/sh is a great example, and probably not what
+  // we're looking for if we say `tea sh` (which is a common thing to do).
+  // So, we should only pass through if the tool we want isn't on the path.
```
mxcl, line 293: This is a good call.

We might want a whitelist.

And we may want `foo` (via magic) to behave differently to `tea foo` as well.

PR #504, opened 2023-04-06
dark magic v0.1.0: pipx

jhheider:


src/hooks/useDarkMagic.ts
```diff
@@ -28,3 +29,17 @@ const npx = async (arg0: string | undefined) => {
 
   return undefined
 }
+
+const pipx = async (arg0: string | undefined) => {
+  const res = await useFetch(`https://pypi.org/pypi/${arg0}/json`)
```
mxcl, line 34: is this how pipx does it?

jhheider: ChatGPT says so.

jhheider: Actually ChatGPT said XML-RPC, but if I wanted JSON that's the variant.

mxcl: I'll cross fingers chat GPT wasn't hallucinating.

mxcl (approved):

jhheider:
It seemed to work ;)

mxcl:
works vs. works exactly like pipx is not necessarily the same thing ofc

PR #505, opened 2023-04-06
dark magic v0.2.0: cargo

jhheider:
this is messy; but all other PMs will mostly work this way. so, we should get it right. but, this works at least. let's fight about better ways.

### DIFF
--- a/src/hooks/useDarkMagic.ts
+++ b/src/hooks/useDarkMagic.ts
@@ -1,11 +1,12 @@
 import { Range } from "semver"
-import { useFetch } from "hooks"
+import { useFetch, usePrefix } from "hooks"
+import { WhichResult } from "types"
 
 export default function useDarkMagic() {
   return { which }
 }
 
-const which = async (arg0: string | undefined) => {
+const which = async (arg0: string | undefined): Promise<WhichResult | undefined> => {
   // Query external providers; return commands to run if found
   const found = await Promise.all([
     npx(arg0),
@@ -25,7 +26,6 @@ const npx = async (arg0: string | undefined) => {
       project: "npmjs.com",
       constraint: new Range("*"),
       shebang: ["npx", "-y"],
-      precmd: undefined
     }
   }
 
@@ -40,7 +40,6 @@ const pipx = async (arg0: string | undefined) => {
       project: "pypa.github.io/pipx",
       constraint: new Range("*"),
       shebang: ["pipx", "run"],
-      precmd: undefined
     }
   }
 
@@ -50,12 +49,21 @@ const pipx = async (arg0: string | undefined) => {
 const cargo = async (arg0: string | undefined) => {
   const res = await useFetch(`https://crates.io/api/v1/crates/${arg0}`)
 
+  const binDir = usePrefix().bin
+
   if (res.status == 200) {
     return {
       project: "rust-lang.org/cargo",
       constraint: new Range("*"),
-      shebang: undefined,
-      precmd: ["cargo", "install", arg0!]
+      explicit: binDir.join(arg0!),
+      precmd: [
+        "cargo",
+        "install",
+        "--quiet",
+        "--root",
+        binDir.parent().string, // installs to {root}/bin
+        arg0!,
+      ]
     }
   }
 
--- a/src/hooks/useDarkMagic.ts
+++ b/src/hooks/useDarkMagic.ts
@@ -10,6 +10,7 @@ const which = async (arg0: string | undefined) => {
   const found = await Promise.all([
     npx(arg0),
     pipx(arg0),
+    cargo(arg0),
   ])
 
   // Return the first non-empty result
@@ -23,7 +24,8 @@ const npx = async (arg0: string | undefined) => {
     return {
       project: "npmjs.com",
       constraint: new Range("*"),
-      shebang: ["npx", "-y"]
+      shebang: ["npx", "-y"],
+      precmd: undefined
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/hooks/useExec.ts
```diff
@@ -239,19 +252,19 @@ export async function which(arg0: string | undefined) {
             // we are being executed via the command not found handler inside a dev-env
             // so let’s use the version that was already calculated for this dev-env
             if ("version" in inenv) {
-              found = {...inenv, constraint: new semver.Range(`=${inenv.version}`), shebang: provider }
+              found = {...inenv, constraint: new semver.Range(`=${inenv.version}`), shebang: provider, precmd: undefined }
             } else {
-              found = {...inenv, shebang: provider }
+              found = {...inenv, shebang: provider, precmd: undefined }
```
mxcl, line 257: not required surely since you defined the type as `precmd?: string`

src/hooks/useDarkMagic.ts
```diff
@@ -37,7 +39,23 @@ const pipx = async (arg0: string | undefined) => {
...
+
+const cargo = async (arg0: string | undefined) => {
+  const res = await useFetch(`https://crates.io/api/v1/crates/${arg0}`)
+
+  if (res.status == 200) {
+    return {
+      project: "rust-lang.org/cargo",
+      constraint: new Range("*"),
+      shebang: undefined,
+      precmd: ["cargo", "install", arg0!]
```
mxcl, line 58: installs to `~/.local/bin/foo`, presumably we should explicitly execute that?

src/hooks/useExec.ts
jhheider: TS was complaining. I'll take another look.

mxcl:
I think we should provide a pkg `cargox` that then behaves like `npx` etc. in that the files are installed to a location controlled by the tool rather than how this works where the executable ends up in `~/.local/bin`

src/hooks/useDarkMagic.ts
jhheider: We've already shown it's not on the path before installing, so that's safer but shouldn't be at issue. *Shouldn't be*...

mxcl:
Or do it internally, but I don't think the executables should end up in a public path.

src/hooks/useDarkMagic.ts
```diff
@@ -37,7 +39,23 @@ const pipx = async (arg0: string | undefined) => {
...
+      shebang: ["pipx", "run"],
+      precmd: undefined
+    }
+  }
+
+  return undefined
+}
+
+const cargo = async (arg0: string | undefined) => {
+  const res = await useFetch(`https://crates.io/api/v1/crates/${arg0}`)
```
mxcl, line 51: Is this working via coincidence? 90% of the time I am sure the executable and pkg are named the same, but it's not guaranteed right?

jhheider: Correct. We'd need some kind of massive data collection to read out the mappings.

jhheider:
We can do that with the --root flag, so, that should be easy.

src/hooks/useDarkMagic.ts
mxcl: massive data collection was in the spec I passed to @owenjames but that can be a next step. It's great to get something already.

I think rebuilds and glibc/linux should come first but will leave that to Owen to figure out.

src/hooks/usePrefix.ts
```diff
@@ -3,10 +3,13 @@ import useConfig from "hooks/useConfig.ts"
 
 export class Prefix extends Path {
   www: Path
+  bin: Path
 
   constructor(prefix: Path) {
     super(prefix)
     this.www = prefix.join("tea.xyz/var/www")
+    // Maybe this should be usr/bin or tmp/bin?
+    this.bin = prefix.join("tea.xyz/bin")
```
mxcl, line 12: I'll change this after merge. I think we may go with `~/.cargo/bin` since it's the default and this works similarly to how `npx` and that work. Probs we should think about having all tea magic stuff cache inside `~/.tea` so users can clear it out. DUNNO YET!

tests/functional/provides.test.ts
```diff
@@ -39,4 +39,12 @@ Deno.test("provides", { sanitizeResources: false, sanitizeOps: false }, async te
     const { run } = await createTestHarness()
     await assertRejects(() => run(["--provides", "armone"]), ExitError, "exiting with code: 0")
   })
+
+  // FIXME: once you _execute_ cargo install, `--provides` returns false,
+  // since it's in the path at that point. But I doubt uninstalling after
+  // run is the right answer.
```
mxcl, line 45: no longer a true comment I believe right?

mxcl (changes requested):
Shouldn't we `usePrefix().bin` somewhere? Or am I missing that

tests/functional/provides.test.ts
jhheider: That's correct; until this moves to .cargo/bin (assuming most cargo users will add it to their path).

jhheider:
> Shouldn't we `usePrefix().bin` somewhere? Or am I missing that

`  const binDir = usePrefix().bin`

useDarkMagic.ts:L52

jhheider:
If it's already installed, it'll show up in the path search before getting here.

mxcl:
> Yes, it invokes, but it doesn't do anything if it's already installed

k well that's fine. If it's quick we can leave it and not mess around.

mxcl (approved):

PR #506, opened 2023-04-06
`PathUtils.ts`: search and manage `$PATH`

jhheider:
if this is out-of-scope, feel free to close. felt like a _common_ path function, on par with `home()`.

### DIFF
--- a/import-map.json
+++ b/import-map.json
@@ -1,6 +1,7 @@
 {
   "imports": {
     "path": "./src/vendor/Path.ts",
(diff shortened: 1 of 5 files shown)


mxcl:
IMO *searching the `PATH`* is not aligned with this library.

However it should be a utility function. Potentially make a new lib `PathUtils.ts` since I agree this is a common thing we (and others) need.

mxcl (approved):
bootiful

PR #507, opened 2023-04-06
dark magic v0.3.0: brew

jhheider:
this uses `Path.findBin()` to test for a `brew` install (we should package brew, probably). If that PR is rejected, then we'll need a new plan.

### DIFF
--- a/src/hooks/useDarkMagic.ts
+++ b/src/hooks/useDarkMagic.ts
@@ -1,6 +1,7 @@
 import { Range } from "semver"
 import { useFetch, usePrefix } from "hooks"
 import { WhichResult } from "types"
+import PathUtils from "path-utils"
 
 export default function useDarkMagic() {
   return { which }
@@ -12,6 +13,7 @@ const which = async (arg0: string | undefined): Promise<WhichResult | undefined>
     npx(arg0),
     pipx(arg0),
     cargo(arg0),
+    brew(arg0),
   ])
 
   // Return the first non-empty result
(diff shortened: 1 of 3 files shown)


src/hooks/useDarkMagic.ts
```diff
@@ -61,3 +63,21 @@ const cargo = async (arg0: string | undefined) => {
 
   return undefined
 }
+
+const brew = async (arg0: string | undefined) => {
+  if (!Path.findBin("brew")) return undefined
+
+  const res = await useFetch(`https://formulae.brew.sh/api/formula/${arg0}.json`)
```
mxcl, line 70: this probably works a reasonable amount of the time. I would however prefer us to run a service that installs all formula and indexes their bin dirs as a longer term project tho.

src/hooks/useDarkMagic.ts
```diff
@@ -52,7 +54,7 @@ const cargo = async (arg0: string | undefined) => {
 
   if (res.status == 200) {
     return {
-      project: "rust-lang.org/cargo",
+      project: "tea.xyz",
```
mxcl, line 57: seems spurious

src/hooks/useDarkMagic.ts
```diff
@@ -61,3 +63,21 @@ const cargo = async (arg0: string | undefined) => {
...
 }
+
+const brew = async (arg0: string | undefined) => {
+  if (!Path.findBin("brew")) return undefined
+
+  const res = await useFetch(`https://formulae.brew.sh/api/formula/${arg0}.json`)
+
+  if (res.status == 200) {
+    return {
+      // FIXME: we should package brew
```
mxcl, line 74: yes we should pkg brew IMO however if we do so bottles must install to `/usr/local` or `/opt/homebrew` to function. So `tea` will need to `sudo` one time on the first instance of using `brew` to change the permissions on those directories.

So as it stands this patch should either wait for brew to pkg’d or check `brew` is otherwise independently installed. Otherwise this is not fully functional.

jhheider: `  if (!Path.findBin("brew")) return undefined`

mxcl: lol I didn't review thoroughly to say the least

mxcl: I see you also said as much in the PR description 🤦🏻‍♂️

PR #515, opened 2023-04-08
fix behavior of `^` in pre-1.0 semver

jhheider:
ref: https://github.com/teaxyz/cli/issues/473#issuecomment-1498925758
ref: https://github.com/jhheider/semverator/commit/7d8ae450a5a2af91256c99273797cf9eed0948c1


mxcl:
hugs

PR #519, opened 2023-04-09
Invoke magic once on fish shell startup

andreiborisov:
`fish` doesn't fire `--on-variable PWD` functions on startup, therefore we need to invoke magic manually once.

Fixes https://github.com/teaxyz/cli/issues/518.


mxcl:
thank you!

PR #520, opened 2023-04-10
tests to ensure resolving package graph msgs do not get printed to stdout

ABevier:


tests/integration.suite.ts
```diff
@@ -86,7 +86,7 @@ const suite = describe({
         // be faster when testing locally
         if ("args" in opts) {
           if (!existing_tea_prefix) {
-            cmd.unshift("--sync", "--silent")
```
mxcl, line 89: like preferably this would be an option to the test function, I added this to make reading the test outputs less tricky due to high noise low signal.

But we'll cope.

mxcl (approved):

PR #530, opened 2023-04-13
Dash dash json

mxcl:
This isn’t all encompassing, but it’s enough for <<REDACTED>>

### DIFF
--- a/src/hooks/useDownload.ts
+++ b/src/hooks/useDownload.ts
@@ -23,7 +23,7 @@ interface RV {
 
 async function internal<T>({ src, headers, logger, dst }: DownloadOptions): Promise<[Path, ReadableStream<Uint8Array> | undefined]>
 {
-  const { isCI, silent } = useConfig();
+  const { isCI, silent, json } = useConfig();
   const { GITHUB_TOKEN } = useEnv();
   logger = isString(logger) ? useLogger(logger) : logger ?? useLogger()
 
@@ -46,8 +46,7 @@ async function internal<T>({ src, headers, logger, dst }: DownloadOptions): Prom
     if (mtime_entry.isFile()) {
       headers["If-Modified-Since"] = await mtime_entry.read()
     }
-    logger.replace(teal('querying'))
-  } else {
+  } else if (!json) {
     logger.replace(teal('downloading'))
   }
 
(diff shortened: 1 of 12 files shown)


src/hooks/useDownload.ts
```diff
@@ -82,26 +85,32 @@ async function internal<T>({ src, headers, logger, dst }: DownloadOptions): Prom
...
+            (logger as Logger).replace(s)
           }
-          (logger as Logger).replace(s)
           controller.enqueue(buf)
       }}))]
     }
   }
   case 304:
-    logger.replace(`cache: ${teal('hit')}`)
+    if (json) {
```
ABevier, line 109: I wonder if there's a good way to create a smarter logger to avoid all these if statements?  So you could instantiate a `Logger` or a `JsonLogger` and it would know how to format stuff.  Although to make that work you'd want to send events to the logger instead of raw text...

Not something for now, but maybe in the future if we get a lot of if/else's

src/hooks/useDownload.ts
```diff
@@ -67,7 +66,11 @@ async function internal<T>({ src, headers, logger, dst }: DownloadOptions): Prom
 
     let txt = teal('downloading')
     if (sz) txt += ` ${gray(pretty_size(sz))}`
-    logger.replace(txt)
+    if (!json) {
+      logger.replace(txt)
+    } else {
+      console.log({status: "downloading"})
```
ABevier, line 72: I think `console.error` is going to be the correct move.

src/hooks/useDownload.ts
mxcl: Yeah indeed. There's certainly a way to make the code cleaner. Both files could just have a `SpecificLogger` class with functions for each step that take the parameters they need to log for the logging type they are doing. Possibly something even more generic but I think it would end up being abstration masturbation.

I considered it but didn't want to spend longer on it and went for the quick win.

src/hooks/useDownload.ts
mxcl: agreed will update.

mxcl:
One issue is that `console.error(obj)` uses arbitrary line breaks (also will truncate, but only for large outputs). I thought it would behave for us but it doesn't seem to always.

Eg while writing the tests I saw that the error object had 3 line breaks while download output seems to be consistently one line.

JSONL requires one line break. And that makes sense as otherwise parsing it requires you to count brackets.

PR #536, opened 2023-04-16
fix semver.Range.toString for `>=0.47<1`

### LINKED ISSUE #535: semver: '>=0.47<1' turns into `~0.47`
Updating `librist` recipe: it was using `mesonbuild.com: ^0.47`, which is becoming `~0.47` (correctly), now. However, `'>=0.47<1'` is also becoming `~0.47`, which is incorrect. Using '>=0.47' for now, with a fixme.

jhheider:
fixes #535

### DIFF
--- a/src/utils/semver.ts
+++ b/src/utils/semver.ts
@@ -183,7 +183,7 @@ export class Range {
       return this.set.map(v => {
         if (!isArray(v)) return `=${v.toString()}`
         const [v1, v2] = v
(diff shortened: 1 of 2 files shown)


mxcl:
Do we need to tag a release?

jhheider:
This is a pretty esoteric bug. Only caught me out because we had a package depending on the "old" behavior.

PR #558, opened 2023-04-29
Handle `package@<version>` format

### LINKED ISSUE #514: `tea python@3` no longer works
How’d this break?

`utils/pkg.ts` can parse eg `node@16` correctly, we used to allow this on the command-line since users expect it and it stopped working.

JrGoodle:
fixes #514

probably a prettier way to do it, but this seems to get the job done

### DIFF
--- a/src/hooks/useExec.ts
+++ b/src/hooks/useExec.ts
@@ -258,7 +258,17 @@ export async function which(arg0: string | undefined) {
             const constraint = new semver.Range(arg0.substring(provider.length))
             found = {...entry, constraint, shebang: [provider] }
           } catch {
-            // not a valid semver range; fallthrough
+            // not a valid semver ^ range; fallthrough
+          }
+
+          // parse with looser requirements iff contains "@" , eg. `node@16`
+          if (arg0.includes("@")) {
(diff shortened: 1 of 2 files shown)


mxcl:
can we demonstrate a usecase?

JrGoodle:
from reading #514 I thought this was a regression.

but one example I can think of could be reading a `Brewfile` for package requirements. it doesn't look like that's implemented currently as one of the possible package sources, but it uses the `package@<version>` syntax.

JrGoodle:
I started [hacking around to add support for Brewfiles](https://github.com/JrGoodle/cli/commit/20853a70edbe73a82351cd04797dfccb61e0c52e) and realized that's not really relevant to this one since it's just about cli arguments

and now I'm wondering if I misunderstood and you were you asking about adding another test?

JrGoodle:
no worries, I just assumed from the way #514 was written that it was a recent regression. makes sense to hold off on it and keep the cli surface area smaller until there's actual demand

mxcl:
I appreciate the PR, I am strict about changes without concrete justification tho. 🤝

PR #559, opened 2023-04-29
Add debug configuration for running cli in vscode

JrGoodle:
I didn't see anything in the docs about how to debug run in vscode so I messed around a bit and got it kind of working. For some reason it adds `deno` as an arg and the symlink check makes it get appended which causes problems. I'm not sure what the current symlink check is for, so could use some guidance there. It seems to hang and never finish, but I haven't looked into that yet.

### DIFF
--- a/src/args.ts
+++ b/src/args.ts
@@ -35,7 +35,10 @@ export function parseArgs(args: string[], arg0: string): [Args, Flags, Error?] {
       ? target.basename()
       : link.basename()
     const match = base.match(/^tea[_+]([^\/]+)$/)
-    if (link.basename() == "deno" && !link.isSymlink()) return  // deno is literally executing our source code
+    const basename = link.basename()
+    // const isSymlink = link.isSymlink()
+    if (basename == "deno") return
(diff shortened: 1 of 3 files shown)


src/args.ts
```diff
@@ -35,7 +35,10 @@ export function parseArgs(args: string[], arg0: string): [Args, Flags, Error?] {
       ? target.basename()
       : link.basename()
     const match = base.match(/^tea[_+]([^\/]+)$/)
-    if (link.basename() == "deno" && !link.isSymlink()) return  // deno is literally executing our source code
+    const basename = link.basename()
+    // const isSymlink = link.isSymlink()
+    if (basename == "deno") return
+    // if (basename == "deno" && !isSymlink) return  // deno is literally executing our source code
```
mxcl, line 41: what's this about?

mxcl: is this necessary to get it debuggable in vscode or?

mxcl:
Closing, reopen if you can help.

PR #589, opened 2023-05-21
use libtea

mxcl:
still a draft as logging was adapted for libtea to an interface that we still need to implement here


mxcl:
a few remaining issues:
* libtea has issues in useSync when using our git
* useLogger stuff is a little off

PR #595, opened 2023-05-26
Typo: remove duplicated --dry-run in --help

beyarkay:
The current verbose help has this list of flags:

```
flags:
[...]
  --dry-run,-n             don’t do anything, just print
[...]
  --dry-run,-n             don’t execute, just print
```

Which has the `--dry-run` flag duplicated. This commit removes the second instance. Since the wording was slightly different for each of the instances, I kept the wording and the ordering as is used for the "less verbose" version of `--help`, which looks like:

```
flags:
  --sync,-S       sync and update environment packages
  --env,-E        inject local environment
  --dry-run,-n    don’t do anything, just print
```


mxcl (approved):

beyarkay:
Thanks for `tea`! I'm really excited to see what it can do for the OSS community 😄

PR #597, opened 2023-06-01
parallelize installs

mxcl:


mxcl:
oh I forgot to restore all the logJSONs… can't merge yet

PR #608, opened 2023-06-07
zsh autocomplete

jhheider:
This works. The places I think it might need love:
- it will be slow. and ever slower. useSync should cache all the provides to the sqlite store for much faster performance.
- ~it wasn't in any way clear to me how to interleave our completions with zsh's defaults. possibly because we don't work like anything else~
  - ~one possibility is to do the path scanning ourselves, so we can just do the whole completion stack. this will also take time, and want for caching.~

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,6 @@
   "deno.config": "deno.jsonc",
   "markdownlint.config": {
     "extends": "./.github/markdownlint.yml"
-  }
+  },
+  "editor.formatOnSaveMode": "modificationsIfAvailable"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,6 +5,5 @@
   "deno.config": "deno.jsonc",
   "markdownlint.config": {
     "extends": "./.github/markdownlint.yml"
-  },
-  "editor.formatOnSaveMode": "modificationsIfAvailable"
+  }
 }
--- a/src/app.magic.ts
+++ b/src/app.magic.ts
@@ -56,23 +56,30 @@ export default function(self: Path, shell?: string) {
         fi
       }
 
-      _tea_completion() {
+      function _tea_completion {
         local completions
 
         if _has_tea_magic; then
           # Call \`tea --complete\` with the current word as the prefix
           completions=($(tea --complete \${words[CURRENT]}))
-          compadd "$completions[@]"
+          if [[ -n "\${completions}" ]]; then
+            _describe "tea" completions
+          fi
         fi
+      }
 
-        # Run the default completions no matter what
-        _complete
+      function _tea_command_names {
+        _tea_completion "$@"
+        _command_names
+      }
+
+      function _tea_completion_wrapper {
+        _tea_command_names "$@"
       }
 
       autoload -Uz compinit
       compinit
-
-      zstyle ':completion:*' completer _tea_completion
+      compdef _tea_completion_wrapper -command-
       `;
   case "elvish":
     // eval ($MAGIC | slurp)
--- a/src/app.magic.ts
+++ b/src/app.magic.ts
@@ -60,8 +60,8 @@ export default function(self: Path, shell?: string) {
         local completions
 
         if _has_tea_magic; then
-          # Call \`tea --autocomplete\` with the current word as the prefix
-          completions=($(tea --autocomplete \${words[CURRENT]}))
+          # Call \`tea --complete\` with the current word as the prefix
+          completions=($(tea --complete \${words[CURRENT]}))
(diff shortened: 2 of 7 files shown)


jhheider:
Those errors are a little thorny. Looks like our zsh package might need:
```yaml
runtime:
  env:
    FPATH: ${{prefix}}/functions
    MODULE_PATH: ${{prefix}}/lib/zsh/{{version.marketing}}
```
or similar...

jhheider:
I believe https://github.com/teaxyz/pantry/pull/2132 will fix this PR, if it's deemed and acceptable solution.

mxcl:
why *is* this dependent on our zsh? zsh comes with mac. Yes, our zsh should work, and linux ofc. But we should be able to add this without fixing zsh and looking at my current schedule I will not be able to fix zsh for weeks.

jhheider:
Probably because of how our tests work. I'll see if I can fix the test.

jhheider:
https://github.com/teaxyz/pantry/pull/2132 should do it.

jhheider:
Confirmed, and `zsh` fixed in the bargain. Ready for critique!

.vscode/settings.json
```diff
@@ -5,5 +5,6 @@
   "deno.config": "deno.jsonc",
   "markdownlint.config": {
     "extends": "./.github/markdownlint.yml"
-  }
+  },
+  "editor.formatOnSaveMode": "modificationsIfAvailable"
```
mxcl, line 9: seems personal

src/app.magic.ts
```diff
@@ -43,15 +43,37 @@ export default function(self: Path, shell?: string) {
...
         fi
       }
-      `
+
+      _tea_completion() {
+        local completions
+
+        if _has_tea_magic; then
+          # Call \`tea --complete\` with the current word as the prefix
+          completions=($(tea --complete \${words[CURRENT]}))
```
mxcl, line 64: even if we use sqlite having to use `tea` will add 80ms to all completion which sucks

src/app.magic.ts
```diff
@@ -43,15 +43,37 @@ export default function(self: Path, shell?: string) {
...
+          compadd "$completions[@]"
+        fi
+
+        # Run the default completions no matter what
+        _complete
```
mxcl, line 69: does this mean we are also running for eg `git foo<TAB>`

mxcl (changes requested):
Tried it and it works!

```
$ py
pygmentize        pyproject-build   python            
pylsp             pytest            python\{\{        
python                               python\{\{\ version.major\ \}\}
python3                              python\{\{\ version.marketing\ \}\}
```

needs some moustache expansion.

```
$ git cl
clang          clear          cluster        
clang++        clippy-driver  clusterdb      
clean -- remove untracked files from working tree
clone -- clone repository into new directory
```

Seems we are completing any time `TAB` is pressed.

mxcl:
Can we have tea completions in teal?

src/app.magic.ts
jhheider: I believe that's a different piece of issue. This is so things in $PATH are also listed appropriately.

.vscode/settings.json
jhheider: can remove

jhheider:
> Can we have tea completions in teal?

worth a try, certainly.

jhheider:
> needs some moustache expansion.
> 

Hm. Might just need to omit mustachioed variants. Otherwise, we'd need to do produce them for every version available...

> Seems we are completing any time `TAB` is pressed.

That we should not do. Probably something I missed about detecting context.

jhheider:
the docs for zsh completion are long but not amazing.

jhheider:
colorizing the results eluded me. i believe everything else is addressed.

mxcl:
> colorizing the results eluded me

was a nice to have, not required. good job trying tho

mxcl:
So trying it out, being able to disntiguish between what is on the system and what is not yet on the system is pretty important IMO.

We either need coloration or to make the results appear completely separately to system results

jhheider:
Zsh completion has a grouping system. I wasn't able to properly group output (complicated, like I said), but supposedly it can be done by tags, then print a header for each section.

I go back and forth about the import. If the point is to be magical, then not showing the difference seems like the right way. If they cared about local vs tea installed, they wouldn't use magic, right?

mxcl:
> I go back and forth about the import. If the point is to be magical, then not showing the difference seems like the right way. If they cared about local vs tea installed, they wouldn't use magic, right?

it's a good point.

Let's just merge and see how it feels for a few weeks I guess.

mxcl (approved):

PR #621, opened 2023-06-28
implement execle with ffi

### LINKED ISSUE #206: Be able to `execvp`
`deno` doesn’t provide it, but we can get it with the [ffi api](https://deno.land/manual@v1.15.0/runtime/ffi_api)

mxcl:
Fixes #206

### DIFF
--- a/src/utils/execve.ts
+++ b/src/utils/execve.ts
@@ -116,7 +116,8 @@ export class CStringArray extends Uint8Array {
         stringsBuffer.subarray(start),
       );
       if (result.read !== result.written) {
-        throw new Error(`not single byte UTF-8 string: ${string}`);
+        //FIXME commented out, but I don't think it matters…?
+        //throw new Error(`not single byte UTF-8 string: ${string}`);
       }
       offset = start + result.written + 1; // Leave null byte
       pointerBuffer[index++] = basePointer + BigInt(start);
--- a/src/utils/execve.ts
+++ b/src/utils/execve.ts
@@ -1,0 +1,90 @@
+import { utils, Path, TeaError } from "tea"
+const { host } = utils
+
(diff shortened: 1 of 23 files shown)


src/utils/execve.ts
```diff
@@ -0,0 +1,83 @@
...
+    case 63: //ENAMETOOLONG:
+    case 8:  //ENOEXEC:
+    case 12: //ENOMEM:
+    case 20: //ENOTDIR:
+    case 26: //ETXTBSY:
+      throw new Error(`execve failed (${libc.symbols.errno})`)
+  }
+}
+
+function arr_to_c(arr: string[]): Deno.PointerValue {
```
lino-levan, line 49: Just wanted to reference a clean way of doing this

https://github.com/aapoalas/libclang_deno/blob/main/lib/utils.ts#L9-L46

lino-levan: Linked implementation doesn't null terminate the array, but it's a pretty trivial change to do so.

mxcl: Unfortunately cannot import directly without adding a bunch of junk:

```
error: Error: Cannot load libclang without LIBCLANG_PATH environment variable
```

So I will copy pasta it ☹️

mxcl: Getting weird issues with this code:

https://github.com/teaxyz/cli/actions/runs/5436798144/jobs/9886901529?pr=621#step:7:17

I'm not really sure what to do about it, the docs for this function are absent.

https://deno.land/api@v1.34.3?s=TextEncoderEncodeIntoResult

Usually it would be a “do another write” type thing but I don’t get the issue so will just revert to the code I had before which—while being less efficient—was good enough and didn't have this issue.

lino-levan: Could always be improved in a follow-up PR.

PR #636, opened 2023-07-10
GPG sign binaries; include pubkey in srcs

### LINKED ISSUE #633: sign tea binaries with the tea.inc. GPG key and attach to the release
This is a precursor to allowing signed binaries from third parties.

We should probs also add the public key to the release tarballs so that the signature can be checked using the key inside the tarball after untarring. Thus being a security loop

jhheider:
Proof of function: https://github.com/teaxyz/cli/actions/runs/5502595335

closes #633

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -104,7 +104,20 @@ jobs:
           APPLE_IDENTITY: ${{ secrets.APPLE_IDENTITY }}
       - name: sanity check
         run: test "$(./tea --version)" = "tea ${{ needs.check.outputs.version }}"
-      - run: tar cJf tea-${{ needs.check.outputs.version }}+${{ matrix.platform.build-id }}.tar.xz tea
+      - name: GPG sign binary
+        run: |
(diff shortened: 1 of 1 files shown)


mxcl:
I'm an idiot, we need signatures for each of the tarballs, not the binary.

Most projects aren’t single binary projects and we want this to be generalized.

jhheider:
Agreed. required as of gpg2 sadly. I mean, the safety of requiring actual passwords, and having a daemon deal with credential management was likely a well-considered decision, but it's overkill for this use case.

mxcl:
have we published our GPG public key in a standard location?

I would have thought `https://tea.xyz/KEY` (but not `KEY`) is a standard of some sort.

jhheider:
we haven't and I just regenerated it because we _had_ been signing with my jacob@tea.xyz key.

the standard is usually to publish to public key servers. In the web3 space, I usually see it prominently mentioned alongside download locations so concerned individuals can check it.

PR #641, opened 2023-07-18
if no pantry and installing, sync

mxcl:
This so brew install mechanism works on first use for user


jhheider:
Is there a good reason not to sync on any attempt at pantry use if the pantry is missing?

PR #643, opened 2023-07-22
Update installaton methods in README.md

marcinjahn:
The previous command did not work on a typical x86 Linux system. The link generated by the script - https://tea.xyz/Linux/x86_64 - was invalid. The updated command is taken directly from https://docs.tea.xyz/getting-started/install-tea

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,9 +230,7 @@ If you prefer, tea is a standalone, cross-platform binary that you can install
 anywhere you want ([releases]). Here’s a handy one-liner:
 
 ```sh
-sudo install -m 755 \
-  <(curl --compressed -LSsf https://tea.xyz/$(uname)/$(uname -m)) \
-  /usr/local/bin/tea
+sh <(curl https://tea.xyz)
 ```
 
 ## Setting up Magic
--- a/README.md
+++ b/README.md
@@ -222,6 +222,14 @@ command not found: node
 
 # Getting Started
 
+The easiest way to install tea is with our installer:
+
+```sh
+sh <(curl https://tea.xyz)
+```
+
+You could also use Homebrew:
+
 ```sh
 brew install teaxyz/pkgs/tea-cli
 ```
@@ -230,7 +238,9 @@ If you prefer, tea is a standalone, cross-platform binary that you can install
 anywhere you want ([releases]). Here’s a handy one-liner:
 
 ```sh
-sh <(curl https://tea.xyz)
+sudo install -m 755 \\
+  <(curl --compressed -LSsf https://tea.xyz/$(uname)/$(uname -m)) \\
+  /usr/local/bin/tea
 ```
 
 ## Setting up Magic


marcinjahn:
Reading the [docs](https://docs.tea.xyz/getting-started/install-tea/without-installer), I realized that you actually want to support both the script I removed in the PR, and the one I added. Then, the proper modification would be to:

- make the script I deleted work
- add the simple installation script that I added to readme

README.md
```diff
@@ -222,6 +222,14 @@ command not found: node
 
 # Getting Started
 
+The easiest way to install tea is with our installer:
+
+```sh
+sh <(curl https://tea.xyz)
```
mxcl, line 228: I actually deliberately removed this recently to encourage people to install via brew as our suggested workflow.

magic can then be added manually, and I intend to improve that.

there's too much worry regarding `curl | sh` installers nowadays and it hinders our adoption.

So yeah, please remove.

README.md
```diff
@@ -230,8 +238,8 @@ If you prefer, tea is a standalone, cross-platform binary that you can install
 anywhere you want ([releases]). Here’s a handy one-liner:
 
 ```sh
-sudo install -m 755 \
-  <(curl --compressed -LSsf https://tea.xyz/$(uname)/$(uname -m)) \
+sudo install -m 755 \\
```
mxcl, line 241: whoops and thanks

mxcl (changes requested):

README.md
marcinjahn: So, would you like to leave only brew installation instruction in the README? I think it'd limit adoption further as well, especially for Linux folks, I guess brew is not that popular on that platform as it is on mac.

mxcl: The instructions also provide a one-liner for installation for people without brew.

so far we've experienced severe reluctance to use our curl installer. So severe I'm just removing it. Use brew or download the standalone `tea` binary.

It's weird since you install brew that way, and I wrote that curl installer too but people are what they are.

README.md
mxcl: seems actually these aren't required.

mxcl:
> The previous command did not work on a typical x86 Linux system. The link generated by the script - https://tea.xyz/Linux/x86_64 - was invalid

@jhheider what we got and what do we need to fix this?

README.md
mxcl: edit: and you say the one-liner doesn't work for linux intel so we should fix that. I added a comment below about that.

jhheider:
```sh
$ curl -OL https://tea.xyz/Linux/x86_64
100   103  100   103    0     0    239      0 --:--:-- --:--:-- --:--:--   241
100 98.1M  100 98.1M    0     0  25.2M      0  0:00:03  0:00:03 --:--:-- 32.4M
$ file x86_64
x86_64: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, with debug_info, not stripped
```

That looks right to me. @marcinjahn did you use `curl`? It only returns anything for `curl`, and is 404 otherwise.

marcinjahn:
> ```shell
> curl -OL https://tea.xyz/Linux/x86_64
> ```

It actually works today, unlike when I originally had tried it.

Anyway, it seems that the proposed changes are not actually needed, I'll close the PR.

PR #646, opened 2023-07-25
tea --magic=install; tea --magic=uninstall

mxcl:
I don't like the ergonomics of `--magic=install` but it was quick n’ easy. Will revise when I have better ideas

### DIFF
--- a/src/app.magic.ts
+++ b/src/app.magic.ts
@@ -243,15 +243,15 @@ export async function install_magic(op: 'install' | 'uninstall') {
 
 function shells(): [Path, string][] {
   const zdotdir = flatmap(Deno.env.get("ZDOTDIR"), Path.abs) ?? Path.home()
-  const fishdir = flatmap(Deno.env.get("XDG_CONFIG_HOME"), Path.abs) ?? Path.home().join(".config")
+  const xdg_dir = flatmap(Deno.env.get("XDG_CONFIG_HOME"), Path.abs) ?? Path.home().join(".config")
 
   const std = (shell: string) => `source <(tea --magic=${shell})  #docs.tea.xyz/magic`
 
   const candidates: [Path, string][] = [
     [zdotdir.join(".zshrc"), std("zsh")],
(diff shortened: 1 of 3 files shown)


jhheider:
`--add` and `--rm`?

src/app.magic.ts
```diff
@@ -176,3 +176,89 @@ export default function(self: Path, shell?: string) {
...
+  }
+
+  if (op == 'uninstall' && !opd_at_least_once) {
+    console.info("magic already not installed")
+  }
+}
+
+function shells(): [Path, string][] {
+  const zdotdir = flatmap(Deno.env.get("ZDOTDIR"), Path.abs) ?? Path.home()
+  const fishdir = flatmap(Deno.env.get("XDG_CONFIG_HOME"), Path.abs) ?? Path.home().join(".config")
```
jhheider, line 246: This is probably better named `const xdgconfigdir` since other compliant tools might use it as the base.

src/app.magic.ts
```diff
@@ -176,3 +176,89 @@ export default function(self: Path, shell?: string) {
...
+
+  const std = (shell: string) => `source <(tea --magic=${shell})  #docs.tea.xyz/magic`
+
+  const candidates: [Path, string][] = [
+    [zdotdir.join(".zshrc"), std("zsh")],
+    [Path.home().join(".bashrc"), 'source /dev/stdin <<<"$(tea --magic=bash)  #docs.tea.xyz/magic'],
+    [Path.home().join(".config/elvish/rc.elv"), std("elvish")],
```
jhheider, line 253: Is this fixed, and not based on XDG_CONFIG_HOME?

jhheider (approved):
Looks sane. Better for shell switchers.

src/app.magic.ts
mxcl: I copied the code from @teaxyz/setup, so it's behaving the same as that but yeah you're right I'm sure. I'll check.

mxcl:
I considered a `--magic` that “magically” determines shell (by testing syntax or magic variables), but it's not worth it probs

PR #653, opened 2023-07-27
Add `--sync` to gitpod and dev docs

abitrolly:
Otherwise, `./tea` fails with

    error: pantry not found: /home/gitpod/.tea/tea.xyz/var/pantry/projects

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -15,5 +15,6 @@ tasks:
       (
         set -e
         .deno/bin/deno task compile
+        ./tea --sync
         echo "./tea is hot"
       )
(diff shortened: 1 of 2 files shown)


mxcl:
tea syncs itself for no pantry in most situations (this is very new behavior, so maybe you have an older tea?)

What command were you running that didn't work?

abitrolly:
@mxcl just opened the project in Gitpod.

```
.deno/bin/deno task compile
./tea node --version
```

mxcl:
maybe I screwed up the code to do this, let me check.

mxcl:
this is fixed with 0.39.5, `--sync` should occur automatically if the pantry is absent.

PR #654, opened 2023-07-27
README.md Add `magic=bash` param

abitrolly:
Otherwise `magic` fizzles

https://github.com/teaxyz/cli/issues/336

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ the install instructions.
 $ which node
 command not found: node
 
-$ source <(tea --magic)
+$ source <(tea --magic=bash)
 
 $ node --version
 tea: installing nodejs.org…


mxcl:
this assumes bash. With no parameters tea detects the shell which should work here.

works for me.

Can you describe what doesn’t work?

abitrolly:
@mxcl again, it doesn't work in Gitpod.

```bash
$ source <(./tea --magic)
error: unsupported shell
$ echo $SHELL
/bin/bash
```

mxcl:
hmm, I am not particularly familiar with gitpod, can you provide a minimal repro case for me? Something is amiss and I cannot figure it out without trying it there.

PR #657, opened 2023-07-27
add $TEA_PREFIX/.local/bin on fish

### LINKED ISSUE #656: Wrong tea bin PATH for config.fish
Not sure where to report this, but it seems the TEA_ONE_LINER generated for `config.fish` is not adding `~/.tea/.local/bin` to my path. Instead it adds `~/.tea/tea.xyz/v*/bin`, and therefore none of the symlinks in `.tea/.local/bin` are available. 

### FISH
```fish
test -d "$HOME/.tea" && "$HOME/.tea/tea.xyz/v*/bin/tea" --magic=fish --silent | source

echo $PATH
/Users/peter/.tea/tea.xyz/v*/bin

which tea
/Users/peter/.tea/tea.xyz/v*/bin/tea

which lsd
(not found)
```

Whereas in zsh the symlinks are available: 

### ZSH
```zsh
test -d "$HOME/.tea" && source <("$HOME/.tea/tea.xyz/v*/bin/tea" --magic=zsh --silent)

echo $PATH
/Users/peter/.tea/.local/bin

which tea
/Users/peter/.tea/.local/bin/tea

which lsd
/Users/peter/.tea/.local/bin/lsd
```

I would submit a PR, but not sure 1) if this is intentional, and 2) if this is an issue with the TEA_ONE_LINER script, or the target script being sourced.

jhheider:
closes #656

### DIFF
--- a/src/app.magic.ts
+++ b/src/app.magic.ts
@@ -59,16 +59,21 @@ export default function(self: Path, shell?: string) {
         "${d}"/tea --env --keep-going --silent --dry-run=w/trace | source
       end
 
+      if not string match -q -r "^$TEA_PREFIX/.local/bin(:|\\$)" $PATH
+        fish_add_path -Ua "$TEA_PREFIX/.local/bin"
+      end
+
       if not string match -q -r "^$HOME/.local/bin(:|\\$)" $PATH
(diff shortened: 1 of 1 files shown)


sporkd (approved):
LGTM 💯 👏

PR #659, opened 2023-07-27
don't rely on .bash_profile to call .bashrc

jhheider:
the mirror of https://github.com/teaxyz/setup/pull/175

ref: https://github.com/teaxyz/setup/issues/172

### DIFF
--- a/src/app.magic.ts
+++ b/src/app.magic.ts
@@ -264,9 +264,12 @@ function shells(): [Path, string][] {
 
   const std = (shell: string) => `source <(tea --magic=${shell})  #docs.tea.xyz/magic`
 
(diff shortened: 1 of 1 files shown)


src/app.magic.ts
```diff
@@ -264,9 +264,12 @@ function shells(): [Path, string][] {
 
   const std = (shell: string) => `source <(tea --magic=${shell})  #docs.tea.xyz/magic`
 
+  const bash = 'test -z "$TEA_HAS_RUN" && source /dev/stdin <<<"$(tea --magic=bash) && export TEA_HAS_RUN=1  #docs.tea.xyz/magic'
```
mxcl, line 267: exporting `TEA_HAS_RUN` to all the user’s terminals is a little gross :/

dunno a good way around it.

notably I was pkging conda.org today and they only add to `.bash_profile`

jhheider: Maybe that's correct? It's possible non login shells inherit from login shells most of the time.

PR #669, opened 2023-07-30
allow vscode to debug tea with arbitrary args

felipecrs:
I was taking a look on https://github.com/teaxyz/cli/issues/664 and I was looking for a simple way to debug tea with something like `tea node --version`.

This does the trick:

(video goes here)


mxcl:
nice! time for me to stop my `console.log` debugging lol

PR #670, opened 2023-07-31
fix pantry not found on first execution

### LINKED ISSUE #664: Automatically `--sync` when needed
When downloading the tea binary and running it for the first time (or when cleaning up stuff like `rm -rf ~/.tea`), tea no longer works with the following error:

```
error: pantry not found: /home/felipecrs/.tea/tea.xyz/var/pantry/projects
```

Running `tea --sync` solves the issue for subsequent executions, until the tea prefix is cleaned up again.

This is somewhat stated in the documentation though:

> `tea --sync` updates pkgs, but you have to call it *at least once*. Our installer does this for you normally

_At https://docs.tea.xyz/getting-started/install-tea/without-installer_

My use case is to bundle Tea with my [dotfiles](https://github.com/felipecrs/dotfiles), but so long this extra step is necessary, I can't simply drop its binary to my dotfiles using [chezmoi](https://www.chezmoi.io/user-guide/include-files-from-elsewhere/). I have this extra step of calling `tea --sync` at least once.

Tea tries to be very smart when resolving dependencies from the environment. Thus, I believe tea could be smarter in this situation as well, not requiring manual intervention to make it work.

Not to mention the error message is definitely not very helpful for beginners who don't even know what tea's pantry is about.

felipecrs:
Fixes #664

### DIFF
--- a/src/app.main.ts
+++ b/src/app.main.ts
@@ -1,7 +1,7 @@
 import { usePrefix, useVirtualEnv, useVersion, usePrint, useConfig, useLogger } from "hooks"
 import { VirtualEnvError, VirtualEnv } from "./hooks/useVirtualEnv.ts"
 import { Verbosity } from "./hooks/useConfig.ts"
-import { Path, utils, semver, hooks } from "tea"
+import { Path, utils, semver, hooks, PantryNotFoundError } from "tea"
 import { basename } from "deno/path/mod.ts"
 import exec, { repl } from "./app.exec.ts"
 import complete from "./app.complete.ts"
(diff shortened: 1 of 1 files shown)


mxcl:
Thank you! However I am fixing this in libtea since I consider it a bug there. Sorry you wasted the time :/

https://github.com/teaxyz/lib/commit/befdbbbd91f7f8f840261cffeaaf871fd2e659d2

mxcl:
will release a fix within couple hours. Waiting on CI

felipecrs:
Your fix is much cleaner, it makes sense. And it's not wasted time at all, I'm in love with Tea. :)

Thank you!

mxcl:
sorry, CI was backed up and I went to sleep, PR up and will merge when it passes

mxcl:
v0.39.5 out

PR #675, opened 2023-08-01
Name binaries with x86_64 rather than non standard x86-64

felipecrs:
This simplifies downloading the binaries from GitHub releases through scripts with `$(uname -m)`.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -75,9 +75,9 @@ jobs:
       matrix:
         platform:
         - os: macos-latest
-          build-id: darwin+x86-64
+          build-id: darwin+x86_64
         - os: ubuntu-latest
-          build-id: linux+x86-64
+          build-id: linux+x86_64
         - os: [self-hosted, macOS, ARM64]
           build-id: darwin+aarch64
         - os: [self-hosted, linux, ARM64]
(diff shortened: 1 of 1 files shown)


felipecrs:
Lol, @what-the-diff looks confused this time.

mxcl:
while I understand the desire for this we elected to use `x86-64` consistently across our suite, partly because underscores are invalid in semantic versions and thus this change breaks the semantic versions in our release tarball filenames.

Now… that's not the end of the world since these are not really meant to be ingested in that way, they are just for humans really.

However it would reduce our consistency, so I guess I'm asking:

What do you want this for exactly?

felipecrs:
PS: I'm aware I can use `https://tea.xyz/$(uname)/$(uname -m)`, but I would prefer downloading directly from GitHub.

felipecrs:
@mxcl I will close this PR, I just wanted to share the message. I suppose you can reopen if you decide in favor of it someday. Cheers!

mxcl:
If we are going to change this then now is the time.

I wonder if we should use the more standard llvm-target-name tho? eg. `linux-unknown-x86_64`

felipecrs:
> I wonder if we should use the more standard llvm-target-name tho? eg. `linux-unknown-x86_64`

Personally I'm all in for reusing existing standards, specially if they make scripting around them easier.

jhheider:
There's quite a few competing standards are this point, sadly. Others also use arm64/x64, both of which would be semver-compliant tokens, if that's the primary concern.

felipecrs:
`amd64` also seems nice, and at least on Debian-based systems, `dpkg --print-architecture` can be used to print it.

jhheider:
I find arm64 and amd64 difficult to distinguish visually on the fly.

PR #703, opened 2023-09-11
Run installed packages silently

davdroman:
I think this would be preferable to the otherwise intrusive logging that may break certain programs or GUIs that expect clean output from tools like e.g. git. And since we can't really install a package without resolving it first, it'd be rather rare for logging to be useful when executing installed packages which have already been resolved.

### DIFF
--- a/src/modes/install.ts
+++ b/src/modes/install.ts
@@ -23,7 +23,7 @@ export default async function(pkgs: PackageRequirement[]) {
         const f = dst.join(program)
         f.write({ text: undent`
           #!/bin/sh
-          exec tea +${utils.pkg.str(pkg)} -- ${program} "$@"
+          exec tea --silent +${utils.pkg.str(pkg)} -- ${program} "$@"
           `}).chmod(0o755)
         console.error('tea: installed:', f)
         n++


davdroman:
@what-the-diff is wrong on this one.

mxcl:
We already prohibit the output if not run in a tty, which we did to handle this situation. Or at least we should if not it's a regression.

Do you have an example where we need to do this so we can study it.

davdroman:
Thanks for the prompt response.

Assuming `.local/bin` is in your `$PATH`:

```sh
$ tea install git

$ tea lazygit
2023/09/11 19:25:25 An error occurred! Please create an issue at: https://github.com/jesseduffield/lazygit/issues

*strconv.NumError strconv.ParseBool: parsing "\x1b[1K\x1b[0Gtea: resolving graph…\x1b[1K\x1b[0G\x1b[1A\x1b[0Jfalse": invalid syntax
/var/folders/y3/yr3bxsy95c3gfwwj61ytbp000000gq/T/4b5ecf61/pkg/app/app.go:55 (0x100a88ca0)
/var/folders/y3/yr3bxsy95c3gfwwj61ytbp000000gq/T/4b5ecf61/pkg/app/entry_point.go:154 (0x100a8a8cc)
/var/folders/y3/yr3bxsy95c3gfwwj61ytbp000000gq/T/4b5ecf61/main.go:23 (0x100a8c128)
/opt/go.dev/v1.20.7/src/runtime/internal/atomic/types.go:194 (0x100449f08)
/opt/go.dev/v1.20.7/src/runtime/asm_arm64.s:1172 (0x10047a624)
```

mxcl:
yeah my bad it regressed, will fix in alpha.2

PR #781, opened 2023-10-06
fix one-liner command

trentlarson:
... since the URL returns a 301 with "Moved Permanently" HTML without the https.

Of course, I'm sure we'd prefer if the DNS got fixed and we don't need to type the full "https://" and we can throw this away.


trentlarson:
Also tried pkgx.dev and that also yields a 301 page.

mxcl:
Thanks, indeed, this was an error

PR #783, opened 2023-10-06
Update pkgx-install.md

j1philli:
Amending pkgx uninstall documentation advising it is coming soon


mxcl:
k well fine, let's fix the markdown lint and probs we should say “just `rm ~/.local/bin/foo`” for now

mxcl:
Thank you, this was indeed worth adding and I was just stubborn.

Issue #788, opened 2023-10-07
`tea uninstall tea` fails


permezel:
Per the FAQ in the docs: `tea uninstall tea` will uninstall.

```
% tea uninstall tea 
tea: command not found: uninstall
```

```
% tea --version
tea 0.39.6
```

```
% tea --sync    
pantries sync’d ⎷
tea: empty pkg env
% tea uninstall    
tea: command not found: uninstall
```

mxcl:
Dupes #752

mxcl:
If you are trying to uninstall `tea` then `rm -rf ~/.tea` and remove the tea binary if placed elsewhere. For `pkgx` it is the same just `~/.pkgx`

PR #790, opened 2023-10-08
pkgx uninstall

### LINKED ISSUE #752: Tea uninstall fails
I'm able to successfully install packages via `tea uninstall xxx` but I am unable to uninstall them as the docs say.
I get the following error:
```bash
× nothing provides: uninstall
│ we haven’t pkgd this yet. can you?
╰─➤ https://docs.tea.xyz/pantry
```

mxcl:
Fixes #752; Refs #788


mxcl:
I like the idea of the scripts themselves containing the uninstall logic since it means no database that may be deleted or corrupted separately.

However we will thus need to uninstall before updating anything that is already installed since we have no other way to return to a clean state first and the `provides` may vary between install invocations.

Basically needs more thought on how it should work

yoroshikun:
To ship this earlier can we "whitelist" packages that have clear install path (simple binary) while others be blocked?
(eg. git, node)

mxcl:
I can prioritize finishing this. I didn’t think it was that important for 2 reasons:

1. Uninstall is much less common an operation than install
2. It’s pretty trivial to do by hand, you just delete the relevant files in one directory, it's not nearly the complexity of a normal package manager uninstall command that realisitically could not be done by hand in 99% of cases.

PR #802, opened 2023-10-11
add instructions to reset dev env to docs

j1philli:
the only thing is the path is macos specific, not sure if there is a better way of saying that in an OS agnostic manner

### DIFF
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -170,6 +170,14 @@ specify just the deps. We support specification as strings, arrays or
 dictionaries so pick the one that feels right to you.
 { endhint %}
 
+## Reset the dev env
+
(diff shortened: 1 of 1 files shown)


j1philli:
well I was just plain mistaken on how this works

mxcl:
> If you need a full reset of a dev environemnt you can remove the dev.pkgx.activated file from the below path:

That is the same as `dev off`.

`dev` automates `env +foo` commands, which is why it outputs what it is calling when you call it.

It builds this by looking at your keyfiles, like `cargo.toml` means `env +rust`.

Calling `env +foo` yourself will add foo but only temporarily since `dev` uses the keyfiles to know what to add.

PR #807, opened 2023-10-13
feat: allow extracting of node+npm from package.json

yoroshikun:
Quick PR for allowing extracting the node and npm version from packages.json

The "pkgx" node version will always have priority, over "engines" use.

I also modified the tests to allow for more than one requirement to be added in one pass, let me know if this is not so good.

#### Possible improvements.
- If pkgx is used it will take over and prevent npm requirement being added if it exists solely in engines. Maybe only check for if a node version is added to the deps.

#### Help
What is the correct way to test the dev function on real projects, do i need to install the build pkgx version and it should use that instead of needing to rebootstrap?


mxcl (approved):
LGTM! Thank you.

> The "pkgx" node version will always have priority, over "engines" use.

I believe constraints are combined so any of them refine the others. Conflicts lead to error eg. one asks for node 16 the other node 18, but if one asks for ^16 and the other ~16.1 then the are combined (in this case the overlap is ~16.1).

I say “believe” rather than know because I'm pretty sure but didn't review the code and wrote it a while back (in mental terms).

> If pkgx is used it will take over and prevent npm requirement being added if it exists solely in engines. Maybe only check for if a node version is added to the deps.

Not sure I understand you here.

> What is the correct way to test the dev function on real projects, do i need to install the build pkgx version and it should use that instead of needing to rebootstrap?

```sh
deno task install
```

Will install your checkout to `/usr/local/bin`. `dev` uses that. Then just step into a `dev`-enabled directory.

yoroshikun:
> If pkgx is used it will take over and prevent npm requirement being added if it exists solely in engines. Maybe only check for if a node version is added to the deps.

@mxcl examples are probs easier.

Currently the code works like this
```json
{
  "engines": {
    "node": "~16.16.1",
    "npm": "~9.7.1"
  }
}
```

Resolved, `nodejs.org~16.16.1,npmjs.com~9.7.1` 

```
{
  "pkgx": "zlib@1.1.1",
  "engines": {
    "node": "~16.16.1",
    "npm": "~9.7.1"
  }
}
```

Resolved `zlib@1.1.1` (only)

since pkgx takes priority.

Based on this I plan to make a follow up PR (or few) to make this a bit more robust at choosing and combining correctly in the packagejson helper.

PR #810, opened 2023-10-15
feat(devenv): detect skaffold.dev using skaffold.yaml file existence

vkumbhar94:


vkumbhar94:
@jhheider @mxcl Pls, have a look when you get some time. Thanks in adv :-)

src/utils/devenv.ts
```diff
@@ -46,6 +46,10 @@ export default async function(dir: Path) {
         pkgs.push({project: "rust-lang.org", constraint})
         await read_YAML_FM(path)  //TODO use dedicated TOML section in preference
         break
+      case "skaffold.yaml":
+        pkgs.push({project: "skaffold.dev", constraint})
+        await read_YAML_FM(path)
```
jhheider, line 51: Instead of using front matter, wouldn't it be advisable to read a `pkgx:` key, similar to package.json?

vkumbhar94: skaffold.yaml has schema validation so adding arbitrary key is not allowed. Let's begin with front matter, I will revisit with skaffold.dev if they allow us to add key.

vkumbhar94: Changed to intelligently find nested dependencies 👍

jhheider: > skaffold.yaml has schema validation so adding arbitrary key is not allowed. 

Ah. Well, you can't fight that.

vkumbhar94: Marking resolved, as I have now removed front matter and finding required deps by reading yaml as user configured.

mxcl (approved):
very nice!

src/utils/devenv.ts
mxcl: also @jhheider in the docs we say “we prefer YAML FM and only use a `pkgx` node for .json since it doesn't allow comments” and I think that's the the right choice even though it's a little more gross.

The solution here though where it adds the deps based on the official YAML is def the most awesome.

Should still also read the YAML FM tho IMO in case user wants to add other deps.

vkumbhar94: In the case of other deps falling outside the skaffold.yaml file's boundary, `pkgx.yaml` at root directory shall be used, that's what I am thinking.

jhheider: > also @jhheider in the docs we say “we prefer YAML FM and only use a `pkgx` node for .json since it doesn't allow comments” and I think that's the the right choice even though it's a little more gross.

this is a sensible response. just feels "funny" to put yaml front matter at the front of... yaml.

PR #823, opened 2023-10-20
+windows

mxcl:


felipecrs:
A suggestion in case you have not thought about it yet:

Perhaps using something like https://github.com/dsherret/dax in brewkit can help minimize the changes needed in pantry regarding the build scripts (assuming you are not going to use MSYS2 or Cygwin to provide native Bash on Windows).

mxcl:
We'll revisit this some other day.

Issue #704, opened 2023-09-11
shell support megaticket


mxcl:
- [] fish
- [] elvish
- [] nushell

amok:
Any plans to implement this (or at least the "fish shell" part) in the nearest future? 🙏 

PS thank you for the awesome and handy tea/pkgx.

mxcl:
I'll give it a go today

mxcl:
may take a while, apparently I can't even write a single fish function it accepts without syntax error

niceEli:
any updates on nushell support

PR #838, opened 2023-10-27
Shellcode for fish

mxcl:
Refs #704

### DIFF
--- a/src/modes/shellcode.ts
+++ b/src/modes/shellcode.ts
@@ -120,18 +120,21 @@ function posixish({tmp, sh, datadir}: {datadir: Path, tmp: Path, sh: string}) {
         echo -e '${dim('^^ type `')}x${dim('` to run that')}' >&2
 
         d="${tmp}/shellcode"
+        u="$d/u.$$"
+        x="$d/x.$$"
         mkdir -p "$d"
 
-        echo "echo -e \\"${blurple('env')} +$1 ${dim('&&')} $@ \\" >&2" > "$d/u.$$"
-        echo "pkgx --internal.use +\\"$1\\"" >> "$d/u.$$"
+        echo "#!${sh}" > "$u"
+        echo "echo -e \\"${blurple('env')} +$1 ${dim('&&')} $@ \\" >&2" >> "$u"
+        echo "pkgx --internal.use +\\"$1\\"" >> "$u"
 
-        echo "#!${sh}" > "$d/x.$$"
-        echo "rm \"$d\"/?.$$" >> "$d/x.$$"
-        echo -n "exec " >> "$d/x.$$"
+        echo "#!${sh}" > "$x"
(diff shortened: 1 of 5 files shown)


src/modes/shellcode.ts
```diff
@@ -203,3 +213,43 @@ function platform_state_default() {
...
+      mkdir -p $d
+
+      echo "echo -e '${blurple("env")} +$cmd ${dim("&&")} $argv' >&2" > "$d/u.$fish_pid"
+      echo "rm \"$tmp\"/shellcode/*.$fish_pid" >> "$d/u.$fish_pid"
+      echo "pkgx --internal.use +$cmd" >> "$d/u.$fish_pid"
+
+      echo "#!/bin/sh" > "$d/x.$fish_pid"
+      echo -n "exec " >> "$d/x.$fish_pid"
+      echo "$argv" >> "$d/x.$fish_pid"  #FIXME needs quoting :/
+      chmod u+x "$d/x.$fish_pid"
```
Lythenas, line 236: If you use a fish script instead of an sh script, you should be able to use something like this:

```fish
echo -n "exec " > "$d/x.$fish_pid"
for arg in $argv
    string escape $arg >> "$d/x.$fish_pid"
    echo -n " " >> "$d/x.$fish_pid"
end
```

And then in the definition of `x` call the script with `fish "${tmp}/shellcode/x.$fish_pid"`. Or use a fish shebang. Not sure which is better.

I also tried to implement fish support here: https://github.com/Lythenas/pkgx/tree/fish-support I basically just translated the posix code, and  since then I think there were changes of the shellcode in pkgx. But feel free to use that as a reference if it is useful. I didn't open a PR because I wasn't really happy with the code (I especially didn't like the --fish flag I added).

mxcl: thank you this is very useful!

joshmedeski:
@mxcl let us know if you need any testers for this!

mxcl:
I think I need to refactor the shellcode and other related pieces before this will be maintainable. I don't use fish (only because in general I stick with defaults so that I know what most of my users deal with on a day to day basis) and as a result it would be easy for it to end up breaking unless the underlying infra is solid.

Also not knowing fish it's hard to write. I am mostly using ChatGPT to translate.

netzhuffle:
@mxcl Do you have anything in specific about how I might be able to help?

joshmedeski:
@mxcl I've got fish scripting experience too, let us know if you want support.

amok:
I'm at your service too, @mxcl. Looking forward to fish support

aybabtme:
This would help contributing to the pantry, as it relies on integration https://github.com/pkgxdev/pantry?tab=readme-ov-file#contributing

mxcl:
For some reason I cannot transfer this issue. Please reopen at https://github.com/pkgxdev/dev

PR #868, opened 2023-11-19
Update terminals.md

motdotla:
Follow any potential redirects

Sorry, super nitpicky! Also, this didn't trip me up for your command. It's just in case someday you or another updates that endpoint to a redirect.

Feel free to close and not merge.


mxcl:
There never will be any redirects. This is by design. When curling a script or binary I feel it is necessary to communicate to the end user that the URL that they are fetching is the final URL without any potentially sneaky redirects. Adding the `-L` flag reduces this certainty.

Thanks all the same.

PR #884, opened 2023-12-08
Updated the docker docs

michaelessiet:


docs/run/anywhere/docker.md
```diff
@@ -49,6 +49,12 @@ options:
 
 ```Dockerfile
 FROM ubuntu
+
+# Install curl in case it isn't already
+RUN apt update
+RUN apt upgrade
+RUN apt install curl
```
mxcl, line 56: do this in one line.

mxcl (changes requested):

docs/run/anywhere/docker.md
```diff
@@ -61,7 +67,14 @@ use a more advanced shell like bash:
...
 FROM ubuntu
+
 SHELL ["/bin/bash", "-c"]
+
+# Install curl in case it isn't already
+RUN apt update
+RUN apt upgrade
+RUN apt install curl
+
 RUN curl https://pkgx.sh | sh
```
jhheider, line 78: ideally, i'd also remove curl at this point so we don't have to worry about library confusion (but it's not strictly necessary)

mxcl: It's a good point, but I prefer the clarity of minimal steps for docs.

We could add that as a “CAUTION” box underneath tho.

PR #893, opened 2023-12-14
Update action node version

eikooc:
Github runners using node16 are being deprecated


mxcl:
This is a fixture. It is never run.

Issue #883, opened 2023-12-07
enable `dev cdk` when cdk.json is present 


dburriss:
Running `cdk` requires a *cdk.json* file. When running `dev` when this file is present, I expect cdk to be enabled for the env, since `cdk` is a known tool to pkgx.

dburriss:
I did some digging and do I understand correctly that if I submitted a PR to [devenv.ts](https://github.com/pkgxdev/pkgx/blob/main/src/utils/devenv.ts#L24) updating the case to look for cdk.json, this should work?

jhheider:
yep, that's the recommended path!

michaelessiet:
Is it safe to close this?

michaelessiet:
@dburriss have you made a PR?

dburriss:
@michaelessiet I did but there are some questions on it around treating

PR #908, opened 2023-12-27
Adds cdk.json to devenv list

dburriss:
This addresses #883 

If _cdk.json_ exists in the folder, `dev` will install the AWS CDK CLI.

image

### DIFF
--- a/fixtures/cdk.json
+++ b/fixtures/cdk.json
@@ -1,0 +1,64 @@
+{
+  "app": "npx ts-node --prefer-ts-exts bin/cdk.ts",
+  "watch": {
+    "include": [
+      "**"
+    ],
+    "exclude": [
+      "README.md",
+      "cdk*.json",
+      "**/*.d.ts",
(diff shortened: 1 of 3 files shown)


mxcl:
>  What sort of test should I write for this? I see the tests for this but don't really understand "supplemental fixture" vs "fixed fixture".

Not sure I remember either. Just add a fixture to `./fixtures` that you can then add to one of the arrays there and get coverage up.

> How would I test this manually? Can I install pkgx from my branch, test it, then flip bach to stable easily?

```sh
$ deno task install
````

Will install your checkout to `/usr/local/bin/pkgx`, the shellcode will use that.

dburriss:
@mxcl 1 issue I did run into was that it was still picking up the 1.1.1 version when I ran `dev`. I just brew uninstalled pkgx and it then used my developer version.

✅ Tests are in and passing. 
✅ The manual test worked (see description).

mxcl (approved):

PR #929, opened 2024-01-16
Trailing colons in env breaks things

mxcl:
Refs https://github.com/pkgxdev/pantry/pull/147

### DIFF
--- a/src/prefab/construct-env.ts
+++ b/src/prefab/construct-env.ts
@@ -27,7 +27,7 @@ export default async function(pkgenv: { installations: Installation[] }) {
     // don’t end with a trailing `:` since that is sometimes interpreted as CWD and can break things
     // instead of `foo:${PATH}` we end up with `foo${PATH:+:PATH}` which is not POSIX but works
     // with all realy shells to avoid the trailing `:`
-    rv[key] = rv[key].replaceAll(new RegExp(`:\\$\{${key}}`, 'g'), `\${${key}:+:$${key}}`)
+    rv[key] = rv[key].replaceAll(new RegExp(`:+\\$\{${key}}$`, 'g'), `\${${key}:+:$${key}}`)
   }
 
   return rv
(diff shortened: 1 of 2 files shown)


src/prefab/construct-env.ts
```diff
@@ -24,6 +24,10 @@ export default async function(pkgenv: { installations: Installation[] }) {
 
   for (const key in rv) {
     rv[key] = rv[key].replaceAll(new RegExp(`\\$${key}\\b`, 'g'), `\${${key}}`)
+    // don’t end with a trailing `:` since that is sometimes interpreted as CWD and can break things
+    // instead of `foo:${PATH}` we end up with `foo${PATH:+:PATH}` which is not POSIX but works
+    // with all realy shells to avoid the trailing `:`
```
jhheider, line 29: `s/realy/really/`

src/prefab/construct-env.ts
```diff
@@ -24,6 +24,10 @@ export default async function(pkgenv: { installations: Installation[] }) {
...
+    rv[key] = rv[key].replaceAll(new RegExp(`:\\$\{${key}}`, 'g'), `\${${key}:+:$${key}}`)
```
jhheider, line 30: this is succinct, but isn't it probably safer to do:
```javascript
  .replaceAll(/::/, ':')
  .replace(/^:/, '')
  .replace(/:$/, '')
```
?

src/prefab/construct-env.ts
mxcl: lol not even sure what I meant to say. “Really” doesn't make sense

jhheider: hm, actually, it's probably `real`.

src/prefab/construct-env.ts
mxcl: I didn't bother with some of these as we construct this env ourselves. The runtime `env` from the package.yml could in theory cause a preceding `:` I guess.

PR #959, opened 2024-01-30
dep on deno 1.40

mxcl:
We cannot apply this yet as deno 1.40 shouts about deprecation warnings in libpkgx that we cannot remove.


jhheider:
what a truly odd design choice. those should be compile-time emissions for 'compiled' output.

PR #969, opened 2024-02-01
use libpkgx bump for sqlite pantry cache

mxcl:


mxcl:
Will release later this week. I have been testing this locally for over a week, but still have concerns that the cache might prove evil for some usages that I will then be annoyed at myself over.

Speed up is 400% though.

jhheider:
all solutions are also problems.

-- someone, probably

PR #979, opened 2024-02-22
Fix dev off when environment is not active

felipecrs:
Which was previously throwing _pkgx_dev_off not found.

And also make sure to pass rm -f as some people alias rm to rm -i, which can come from the popular OMZ common-aliases: https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/common-aliases#file-handling

### DIFF
--- a/src/modes/internal.activate.ts
+++ b/src/modes/internal.activate.ts
@@ -95,7 +95,7 @@ export default async function(dir: Path, { powder, ...opts }: { powder: PackageR
       env ${off_string}
 
       if [ "$1" != --shy ]; then
-        rm "${persistence}"
+        rm -f "${persistence}"
       fi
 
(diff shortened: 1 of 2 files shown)


jhheider:
> And also make sure to pass rm -f as some people alias rm to rm -i, which can come from the popular OMZ common-aliases: https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/common-aliases#file-handling

could also just be explicit about `/bin/rm` to avoid dealing with aliases.

felipecrs:
> could also just be explicit about `/bin/rm` to avoid dealing with aliases.

Right, or better yet `command rm`. But I find `rm -f` nicer.

Of course, let me know if you want me to change it.

mxcl:
We should be explicit with `/bin/rm` will patch that in after.

PR #989, opened 2024-03-14
fix `Assertion failed: pkgx: env is being duped` when multiple unicodes in env

### LINKED ISSUE #988: `mkenv` invariant assertion invalid with duplicate `unicode.org` packages.=
Eeeeenteresting. I can guess exactly why:

The error comes from here: https://github.com/pkgxdev/pkgx/blob/104affb636a128b9e9c6b7fb2f07f2f0ea17f128/src/prefab/construct-env.ts#L38-L40

And as of this PR: https://github.com/pkgxdev/libpkgx/pull/61, on January 27th, `libpkgx` allows multiple versions of `unicode` in the environment, _specifically_ to deal with `postgresql` requiring v73 and most of our other packages built for v71. So, it is both the case that the assertion is correct, and the behavior is also correct.

Due to that PR, the assertion is outdated, since we can, now, have more distinct packages than unique packages. So, good find.

_Originally posted by @jhheider in https://github.com/pkgxdev/pkgx/issues/984#issuecomment-1998325406_

jhheider:
closes #988

### DIFF
--- a/src/prefab/construct-env.ts
+++ b/src/prefab/construct-env.ts
@@ -36,7 +36,7 @@ export default async function(pkgenv: { installations: Installation[] }) {
 
 ///////////////////////// reworked from useShellEnv needs porting back to libpkgx
 async function mkenv({installations}: {installations: Installation[]}) {
-  const projects = new Set(installations.map(x => x.pkg.project))
+  const projects = new Set(installations.map(x => `${x.pkg.project}@${x.pkg.version}`))
   console.assert(projects.size == installations.length, "pkgx: env is being duped")
 
   const common_vars: Record<string, OrderedSet<string>> = {}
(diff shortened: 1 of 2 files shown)


src/prefab/construct-env.ts
```diff
@@ -36,7 +36,7 @@ export default async function(pkgenv: { installations: Installation[] }) {
 
 ///////////////////////// reworked from useShellEnv needs porting back to libpkgx
 async function mkenv({installations}: {installations: Installation[]}) {
-  const projects = new Set(installations.map(x => x.pkg.project))
+  const projects = new Set(installations.map(x => `${x.pkg.project}@${x.pkg.version}`))
```
mxcl, line 39: Can't remember why we added this anyway.

mxcl (approved):

src/prefab/construct-env.ts
jhheider: it's been a while, but i believe it's because we can allow multiple unicode lib versions in the env simultaneously, and this check trips if there are two or more (since they map to the same name).

mxcl: I seem to recall there was an issue if an env had multiple of the same exact thing but can't remember.

jhheider: i _do_ recall at one point early on we would sometimes get the pkg@ver duplicated; perhaps this was to ensure that bug never returned.

Issue #998, opened 2024-04-24
pkgx install should overwrite existing versions


felipecrs:
Let's say I previously had:

```console
pkgx install java@17
```

Then later:

```console
$ pkgx install java@11
pkgx: already installed: java (openjdk.org^17)

$ echo $?
0
```

Here is what I think:

1. `pkgx install` should uninstall if previously installed and then install again the new version. Just like `sudo apt-get install docker-ce-cli=5:25.0.3-1~ubuntu.22.04~jammy`.

1. If 1 is too hard to be implemented, the least pkgx should do is to return a non-zero code when another version of the package is already installed.

This happened in my CI/CD, where I ship an environment with `pkgx install java@17`, but some services needed to use `java@11`.

Their pipelines were configured to run `pkgx install java@11` but since a succesful exit code was given, the application was still being built with java@17.

jhheider:
yeah, the non-zero exit code feels incorrect.

as far as your CI/CD goes, even something as simple as: `rm -r ~/.local/bin || true` should be sufficient, if you need to use the `install` functionality, rather than either an environment or shebang-style invocation.

felipecrs:
> as far as your CI/CD goes, even something as simple as: `rm -r ~/.local/bin || true` should be sufficient, if you need to use the `install` functionality, rather than either an environment or shebang-style invocation.

Yeah. For now I do `pkgx uninstall java` before `pkgx install`.

PR #999, opened 2024-04-24
Error when a different version is already installed

felipecrs:
This is just a quick take on #998.

Ideally, pkgx should be able to handle this case by first uninstalling the existing package and then installing the desired one, which would in turn make scripting with pkgx a lot easier.


felipecrs:
image

PR #1000, opened 2024-04-27
Escape spaces in stubs for uninstallation

felipecrs:
I noticed this problem when trying to uninstall Python:

image

Obviously, the source problem is something else that should be fixed. But nevertheless, the stubs should be protected. There may exist some real program name with spaces as well, for example.

## Before

image

## After

image

### DIFF
--- a/src/modes/install.ts
+++ b/src/modes/install.ts
@@ -45,7 +45,7 @@ export default async function(pkgs: PackageRequirement[]) {
             ${exec}
           else
             cd "$(dirname "$0")"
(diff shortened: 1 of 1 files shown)


felipecrs:
The `-f` prevents errors like these:

```console
$ pkgx uninstall jfrog
rm: cannot remove 'jf': No such file or directory
```

felipecrs:
@mxcl, please consider this quick fix for the next release. :)

mxcl (approved):

PR #1008, opened 2024-05-11
feat(install): add support for `--unsafe` install flag

### LINKED ISSUE #991: Programs installed by pkgx needs to be quicker
It would be great if the programs installed by pkgx were quicker. I use `pkgx install` in my devcontainers to ensure the same environment is used by both developers and CI/CD (Jenkins).

However, I'm noticing a very significant increase in the time it takes to execute the same scripts when using pkgx installed programs. One of my pipelines that usually takes 10 minutes started to take 15 minutes after switching to pkgx.

For a comparison:

```console
> pkgx install node@20.12.1 && ~/.local/bin/node --version

# using pkgx's shim to execute node
> time ~/.local/bin/node --version
v20.12.1
~/.local/bin/node --version  0.37s user 0.30s system 128% cpu 0.519 total

# pkgx's node without any shim
> env +node@20.12.1

> time node --version
v20.12.1
node --version  0.00s user 0.00s system 59% cpu 0.006 total

> volta install node@20.12.1 && ~/.volta/bin/node --version

# using volta's shim to execute node
> time ~/.volta/bin/node --version
v20.12.1
~/.volta/bin/node --version  0.00s user 0.00s system 30% cpu 0.006 total

# volta's node without any shim
> time $(volta which node) --version
v20.12.1
$(volta which node) --version  0.00s user 0.00s system 11% cpu 0.004 total
```

As you can see, pkgx's shim seems to add 0.5-1s on every invocation. Which is reasonable for day to day use but is unnaceptable for some scripts and environments.

Another comparison is when using aqua-proxy (made with Go), which seems to add around 50ms to each invocation.

Volta's shim is made with Rust, and it is extremely fast. It even looks for a `package.json` in the current and up folders to find out the proper node version to use. And still, extremely fast.

Maybe shipping some kind of pkgx-shim binary made with Rust would be the best option in this case.

### LINKED ISSUE #997: Can we get more scalable programs on PKGX?
The projects on PKGX are not fast enough.

rustdevbtw:
closes #997
closes #991

This PR adds support for the `--unsafe` install flag to PKGX, which, as proposed by @jhheider and @felipecrs (thanks!), creates additional env stubs for a package on the first run, making subsequent runs make use of the already installed binaries, thus making it *faster*[1].

## The stub
As for the stub, it makes it a *bit* different, for instance, here's a "safe" `node` wrapper (installed normally):
```sh
#!/bin/sh
if [ "$PKGX_UNINSTALL" != 1 ]; then
  exec pkgx +nodejs.org -- node "$@"
else
  cd "$(dirname "$0")"
  rm node && echo "uninstalled: nodejs.org" >&2
fi
```
And here's the "unsafe" (as in experimental) `node` wrapper:
```diff
 #!/bin/sh
+#MANAGED BY PKGX
 if [ "$PKGX_UNINSTALL" != 1 ]; then
-  exec pkgx +nodejs.org -- node "$@"
+  ARGS="$*"
+  ENV_FILE="${XDG_CACHE_DIR:-$HOME/.cache}/pkgx/envs/nodejs.org.env"
+  PKGX_DIR="${PKGX_DIR:-$HOME/.pkgx}"
+
+  pkgx_resolve() {
+    mkdir -p "$(dirname "$ENV_FILE")"
+    pkgx +nodejs.org 1>"$ENV_FILE"
+    run
+  }
+  run() {
+    if test -e "$ENV_FILE" && test -e "$PKGX_DIR/nodejs.org/v*/bin/node"; then
+      set -a
+      # shellcheck source=$ENV_FILE
+      . "$ENV_FILE"
+      set +a
+      exec "$PKGX_DIR/nodejs.org/v*/bin/node" "$ARGS"
+    else
+      pkgx_resolve
+    fi
+  }
+  run
 else
-  cd "$(dirname "$0")"
+  cd "$(dirname "$0")" || exit
   rm emacs && echo "uninstalled: nodejs.org" >&2
 fi
```
> Edit: this version of the stub is outdated, and doesn't match the actual one implemented. Please refer to the code instead.

## Additional changes
Additionally, this PR also makes the following behavioural changes to PKGX:
1. `pkgx uninstall` now also removes the env stub of the package
2. `pkgx install` now checks for *both* `exec pkgx` (old behaviour) and `#MANAGED BY PKGX` (comment) to determine whether that file was created by PKGX.

## Usage
To do an unsafe installation, there are two options:
1. Use the `--unsafe` flag (e.g `pkgx install --unsafe node`)
2. Use the `PKGX_UNSAFE_INSTALL` environmental variable.

In the presence of either, the `--unsafe` flag takes precedence.

## Caveats
This approach doesn't come without any caveat. It's necessary to document them upon having such a functionality.
As per my knowledge, these are the known issues:
1. If the cache dir doesn't exist (*after* installing the package), it fails.
> Edit: it has been fixed
2. If the envs need updating at a later point, that has to be done separately.
3. If a certain dependency is removed, it doesn't properly check for that, and would error.
4. These stubs aren't 100% managed by PKGX (unless you remove the env, when they'll be updated by PKGX)

Most (if not all) of them error correctly and offer (potentially) helpful error messages.  

[1]
benchmarking node, the `node.unsafe` is newer one, and the `node.safe` is the normal one
benchmarking emacs, the `emacs.unsafe` is the newer one, and the `emacs.safe` is the normal one


rustdevbtw:
Here is another benchmark, comparing `node.unsafe` (installed through `--unsafe`), `node.safe` (normally installed), `pkgx node`, and the system-installed (in this case, installed from Arch repos) `node`:
benchmarking node  

I believe (though I could be wrong) that the 1.1x diff is because of the shell overhead, and overhead of checking files.

rustdevbtw:
Change the `mkdir -p` (on the first run, or when the cache doesn't exist) to do that on `$(dirname ~/.cache/pkgx/envs/gnu.org/emacs.env)` so it creates the `gnu.org` dir also (the install also takes care of this, but if the cache is missing, it'll cause error, as mentioned in 1st caveat). It handles the first caveat.

src/modes/install.ts
```diff
@@ -39,7 +46,30 @@ export default async function(pkgs: PackageRequirement[]) {
         if (program.includes("{{")) continue
 
         const pkgstr = utils.pkg.str(pkg)
-        const exec = `exec pkgx +${pkgstr} -- ${program} "$@"`
+        if (UNSAFE) {
+          const parts = pkgstr.split("/")
+          parts.pop()
+          await Deno.mkdir(Path.home().join(`.cache/pkgx/envs/${parts.join("/")}`).toString(), {recursive: true})
```
jhheider, line 52: this should almost certainly be
```typescript
  config.cache.join(`pkgx/envs/${parts.join("/"}`).mkdir('p')
```
to use our standard `Path` library and `Config` struct.

src/modes/install.ts
```diff
@@ -39,7 +46,30 @@ export default async function(pkgs: PackageRequirement[]) {
...
+            pkgx_resolve() {
+              mkdir -p "$(dirname \\"$\{XDG_CACHE_DIR:-$HOME/.cache\}/pkgx/envs/${pkgstr}.env\\")"
+              pkgx +${pkgstr} 1>"$\{XDG_CACHE_DIR:-$HOME/.cache\}/pkgx/envs/${pkgstr}.env"
+              run
+            }
+            run() {
+              if [[ -e "$\{XDG_CACHE_DIR:-$HOME/.cache\}/pkgx/envs/${pkgstr}.env" && -e "$\{PKGX_HOME:-$HOME/.pkgx\}/${pkgstr}/v*/bin/${program}" ]]; then
+                set -a
+                source "$\{XDG_CACHE_DIR:-$HOME/.cache\}/pkgx/envs/${pkgstr}.env"
+                exec "$\{PKGX_HOME:-$HOME/.pkgx\}/${pkgstr}/v*/bin/${program}" "$ARGS"
```
jhheider, line 66: you should probably follow this with a `set +a` for proper termination.

src/modes/install.ts
```diff
@@ -39,7 +46,30 @@ export default async function(pkgs: PackageRequirement[]) {
...
 
         const pkgstr = utils.pkg.str(pkg)
-        const exec = `exec pkgx +${pkgstr} -- ${program} "$@"`
+        if (UNSAFE) {
+          const parts = pkgstr.split("/")
+          parts.pop()
+          await Deno.mkdir(Path.home().join(`.cache/pkgx/envs/${parts.join("/")}`).toString(), {recursive: true})
+        }
+        //FIXME: doing `set -a` clears the args env
+        const exec = UNSAFE ? undent`
```
jhheider, line 55: i'd probably alias the substitutions so you don't have to keep doing them:
```shell
  ENV_FILE=\"$\{XDG_CACHE_DIR:-$HOME/.cache\}/pkgx/envs/${pkgstr}.env\"
  PKGX_DIR=$\{PKGX_DIR:-$HOME/.pkgx\}
```

src/modes/install.ts
```diff
@@ -65,7 +95,7 @@ export default async function(pkgs: PackageRequirement[]) {
             if (done) {
               throw new PkgxError(`${f} already exists and is not a pkgx installation`)
             }
-            const found = value.match(/^\s*exec pkgx \+([^ ]+)/)?.[1]
+            const found = value.match(/^\s*pkgx \+([^ ]+)/)?.[1]
```
jhheider, line 98: this might be dangerous, since it would catch a script with `#!/usr/bin/env pkgx +python ...`

src/modes/uninstall.ts
```diff
@@ -11,6 +11,15 @@ export default async function(pkgspecs: string[]) {
 async function uninstall(prefix: Path, pkgs: PackageRequirement[]) {
   for (const pkg of pkgs) {
     const programs = await hooks.usePantry().project(pkg).provides()
+    const pkgstr = utils.pkg.str(pkg)
+    const parts = pkgstr.split("/")
+    parts.pop()
+    //FIXME: it removes the dir successfully. however, it still complains that it didn't delete that
+    try {
+      await Deno.remove(Path.home().join(`.cache/pkgx/envs/${parts}`).toString(), {recursive: true})
```
jhheider, line 19: This should also use `Config` and `Path`.

src/parse-args.ts
```diff
@@ -89,6 +91,9 @@ export default function(input: string[]): Args {
       case 'update':
         flags.update = true
         break
+      case 'unsafe':
+        flags.unsafe = true
```
jhheider, line 95: you could simplify ENV handling by moving it here; i don't know if we do that otherwise (as with, say, verbosity or debug).

src/modes/install.ts
rustdevbtw: Is that `'p'` argument the option? (as in `mkdir -p`)

src/modes/install.ts
rustdevbtw: So, after sourcing, it should `set +a` (to preserve env)?

src/modes/install.ts
rustdevbtw: Thought that but forgot at the last minute. Gonna add.

src/modes/install.ts
jhheider: yep: https://github.com/pkgxdev/libpkgx/blob/2aec73a0c16222d2c1a4eddb96acc867dddaa191/src/utils/Path.ts#L351-L357

src/modes/install.ts
rustdevbtw: Is there some other way of detecting both normal and unsafe packages? We can use special regex to match maybe both `exec pkgx` OR some other unsafe-specific (non-`pkgx`) match?

src/modes/install.ts
rustdevbtw: k

rustdevbtw: From where that `config` var comes though? (Sorry for being an idiot, lol)

rustdevbtw: Is that from `useConfig()`?

jhheider: Not at all. it should be from `const config = useConfig()` where `Config` comes from https://github.com/pkgxdev/libpkgx/blob/main/src/hooks/useConfig.ts

src/modes/install.ts
jhheider: i _believe_ that will return it to default behavior, yes.

src/modes/install.ts
jhheider: in this instance. at least, you could search for `${pkgstr}.env`. that should be unique to your shim.

src/modes/install.ts
rustdevbtw: That's what I meant:
```ts
const config = useConfig();
config.cache.join(`envs/${parts.join("/")}/`).mkdir("p");
```

src/parse-args.ts
jhheider: worth starting a discussion about, definitely. experience teaches us that there are sometimes unwritten design principles at play; though, in general, providing for either flag or env control allows a wider variety of use-cases.

rustdevbtw:
@jhheider It seems to be okay, but I'm not in PC rn (it's 1:02AM here lol). Can you try it  now please, if possible?

src/modes/install.ts
jhheider: lol, `not at all` was in re: your idiot remark. we crossed each other.

rustdevbtw: xD

src/modes/install.ts
rustdevbtw: What if instead, unsafe installations contain a specific line (like `PKGX_UNSAFE=1`) and check that? Unlike safe (normal) installations, we have full control over how the scripts are gonna be, and there's no backwards compatibility requirement.

jhheider:
> @jhheider It seems to be okay, but I'm not in PC rn (it's 1:02AM here lol). Can you try it now please, if possible?

bugs in the script, but i can add a commit to fix them.

jhheider:
but the speed looks good:
```
pkgx hyperfine "node --version"
Benchmark 1: node --version
  Time (mean ± σ):      26.4 ms ±   2.1 ms    [User: 18.8 ms, System: 3.5 ms]
  Range (min … max):    25.0 ms …  46.3 ms    103 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```

rustdevbtw:
> > @jhheider It seems to be okay, but I'm not in PC rn (it's 1:02AM here lol). Can you try it now please, if possible?
> 
> bugs in the script, but i can add a commit to fix them.

Thanks. Sucks to be without laptop/PC, lol. Anyways, thanks again.

rustdevbtw:
> but the speed looks good:
> ```
> pkgx hyperfine "node --version"
> Benchmark 1: node --version
>   Time (mean ± σ):      26.4 ms ±   2.1 ms    [User: 18.8 ms, System: 3.5 ms]
>   Range (min … max):    25.0 ms …  46.3 ms    103 runs
> 
>   Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
> ```

Is that right after doing the install, or after doing an initial run? It caches only at the first run, so subsequent runs should be faster than the first one. And the first one might interfere with the rest.

src/modes/install.ts
```diff
@@ -39,15 +46,52 @@ export default async function(pkgs: PackageRequirement[]) {
...
-            rm ${programs.join(' ')} && echo "uninstalled: ${pkgstr}" >&2
-          fi`
-        const f = dst.mkdir('p').join(program)
+
+        let script = ""
+
+        if (UNSAFE) {
+          const config = hooks.useConfig()
+          const parts = pkgstr.split("/")
+          parts.pop()
```
jhheider, line 55: this is probably better done with slice, but that's just a style choice

src/modes/install.ts
jhheider: i meant that the script should contain the string `${pkgstr}.env`, since they source it.

rustdevbtw: Yeah, that could be done. Probably with regex as well.

src/parse-args.ts
rustdevbtw: What about `PKGX_UNSAFE_INSTAL=0`? It means false, so we shouldn't consider it to be set. And if it's this, we should check for the other.

src/modes/install.ts
rustdevbtw: Done

src/parse-args.ts
jhheider: I would think if either --unsafe or PKGX_UNSAFE_INSTALL != 0, then we do an unsafe install. PKGX_UNSAFE_INSTALL=0 should basically do nothing.

rustdevbtw:
You mean, the env should take priority?

Then, it needs to be like:
```ts
function is_unsafe(unsafe: boolean): boolean {
    const env = Deno.env.get("PKGX_UNSAFE_INSTALL");
    if (env) {
        // If present, it takes priority
        return parseInt(env) ? true : false;
    } else {
       // Otherwise, check flag
       return unsafe;
    }
```

src/parse-args.ts
rustdevbtw: You mean, the env should take priority?

Then, it needs to be like:
```ts
function is_unsafe(unsafe: boolean): boolean {
    const env = Deno.env.get("PKGX_UNSAFE_INSTALL");
    if (env) {
        // If present, it takes priority
        return parseInt(env) ? true : false;
    } else {
       // Otherwise, check flag
       return unsafe;
    }
```

jhheider:
> You mean, the env should take priority?

No, I mean if either is set, it should be on. Which is the way you have it, I believe. Setting the envvar to zero shouldn't do anything.

rustdevbtw:
> > You mean, the env should take priority?
> 
> No, I mean if either is set, it should be on. Which is the way you have it, I believe. Setting the envvar to zero shouldn't do anything.

By shouldn't do anything, do you mean it should ignore that? So, if that's set, regardless of its value, it should do an unsafe install?

rustdevbtw:
> > > You mean, the env should take priority?
> > 
> > No, I mean if either is set, it should be on. Which is the way you have it, I believe. Setting the envvar to zero shouldn't do anything.
> 
> By shouldn't do anything, do you mean it should ignore that? So, if that's set, regardless of its value, it should do an unsafe install?

If yes, it gets simpler:
```ts
function is_unsafe(unsafe: boolean): boolean {
    return Deno.env.get("PKGX_UNSAFE_INSTALL") || unsafe;
}
```

jhheider:
> So, if that's set, regardless of its value, it should do an unsafe install?

No. If it's non-zero, or the --unsafe flag is passed, it should do an unsafe install. If the envvar is zero, it has no effect one way or the other.

rustdevbtw:
The default behaviour (without any env or flag) is to install normally. So,
if the env is set to 0, it should do that? Sorry for sounding like an idiot.

jhheider:
> The default behaviour (without any env or flag) is to install normally. So, if the env is set to 0, it should do that?

Exactly. Setting it 0 is the same as not setting it. So, it doesn't affect behavior.

rustdevbtw:
> > The default behaviour (without any env or flag) is to install normally. So, if the env is set to 0, it should do that?
> 
> Exactly. Setting it 0 is the same as not setting it. So, it doesn't affect behavior.
> 
> If it's set to 0 and the flag is passed it's unsafe. If it's 0 and no flag, then safe. Set to 0 is a no-op.

So, this:
```ts
function is_unsafe(unsafe: boolean): boolean {
    const env = parseInt(Deno.env.get("PKGX_UNSAFE_INSTALL") || "0") ? true : false;
    return env || unsafe;
}
```

rustdevbtw:
@jhheider implemented it, is it okay? If so, can you please test that the env works as expected? (It probably does, but still a test is a good idea)

rustdevbtw:
So @jhheider anything else that's left? (Maybe check if it's by PKGX part? We can improve that)

Also @mxcl can you please share thoughts?

jhheider:
looks like good work

felipecrs:
> 2. pkgx install now checks for pkgx (instead of exec pkgx) in a script to consider it as installed by PKGX (because unsafe installations don't use exec pkgx)

I've got to say this approach is just as not future-proof as the previous. Perhaps we should fix it for good by adding some known comment to the stubs? Like:

```bash
#!/bin/sh
# managed by pkgx
```

felipecrs:
> as proposed by @jhheider (thanks!)

Guys, I don't want to be silly, but I had [proposed this on Apr 11](https://github.com/pkgxdev/pkgx/issues/991#issuecomment-2048921757). @jhheider is there a chance you missed my comment and later came up with the same proposal?

rustdevbtw:
> > 2. pkgx install now checks for pkgx (instead of exec pkgx) in a script to consider it as installed by PKGX (because unsafe installations don't use exec pkgx)
> 
> I've got to say this approach is just as not future-proof as the previous. Perhaps we should fix it for good by adding some known comment to the stubs? Like:
> 
> ```bash
> #!/bin/sh
> # managed by pkgx
> ```

That is possible, but it'll break compatibility with the stubs installed by older versions. PKGX wouldn't detect them as installed by PKGX if you completely move to this approach. As for unsafe install, I've had the idea to see and check for .env files (generated for unsafe installations only). That's not as reliable, though, because it would fail if the cache was somehow deleted, or not yet generated.

felipecrs:
> That is possible, but it'll break compatibility with the stubs installed by older versions. PKGX wouldn't detect them as installed by PKGX if we completely move to this approach.

Perhaps we can keep searching for `pkgx` while still in pkgx 1.x, and maybe drop support for it in pkgx 2.x. Anyway, this can be done in another PR, no need to have it done here.

rustdevbtw:
maybe i read that proposal before coming up with this PR, and forgot
about it before doing it. it is certainly inspired by your proposal.
so, i think you deserve the credit!

rustdevbtw:
> > That is possible, but it'll break compatibility with the stubs installed by older versions. PKGX wouldn't detect them as installed by PKGX if we completely move to this approach.
> 
> Perhaps we can keep searching for `pkgx` while still in pkgx 1.x, and maybe drop support for it in pkgx 2.x. Anyway, this can be done in another PR, no need to have it done here.

That sounds like a good option. What do you think @jhheider?

jhheider:
sure, sounds like a good idea.

rustdevbtw:
does ***@***.*** BY PKGX@` sounds good for that comment? (it's unique enough
to not cause conflicts)

jhheider:
go for it!

rustdevbtw:
> go for it!

done!

rustdevbtw:
Also, updated the example to reflect the current `diff` (as of now)

mxcl:
Nice feature. Additive so easy merge.

PR #1037, opened 2024-09-12
fix regression in construct-env.ts

### LINKED ISSUE #1036: strange behaviors with 1.2.0
These are different:

CMAKE_INCLUDE_PATH doesn't seem to be being set: https://github.com/pkgxdev/pantry/actions/runs/10776133971/job/29882224768 (1 example; a fix for another was explicit paths: https://github.com/pkgxdev/pantry/pull/7236/files

LD_LIBRARY_PATH (and path, weirdly) don't seem to include runtime deps with brewkit test? maybe? or the output is strange from brewkit is strange: https://github.com/pkgxdev/pantry/actions/runs/10820659200/job/30021167540. This should show php.net in PATH (though it clearly works on the command), and LD_LIBRARY_PATH doesn't seem to include gnu.org/gcc/libstdcxx. weirdly, laravel (#!/usr/bin/env php) works fine, but `php artisan` and `./artisan`, same bang-path, have GLIBCXX errors.

note that the second appears to be the same issue you saw in CI/CD with nodejs.

jhheider:
closes #1036

### DIFF
--- a/src/prefab/construct-env.ts
+++ b/src/prefab/construct-env.ts
@@ -37,7 +37,11 @@ export default async function(pkgenv: { installations: Installation[] }) {
 ///////////////////////// reworked from useShellEnv needs porting back to libpkgx
 async function mkenv({installations}: {installations: Installation[]}) {
   const projects = new Set(installations.map(x => `${x.pkg.project}@${x.pkg.version}`))
-  console.assert(projects.size == installations.length, "pkgx: env is being duped")
+
+  // we need to do this as well, since we allow multiple versions of e.g. unicode.org.
+  // however, if we *just* use this as `projects`, then tests below like `projects.has('cmake.org')` will fail
+  const projects_with_versions = new Set(installations.map(x => `${x.pkg.project}@${x.pkg.version}`))
+  console.assert(projects_with_versions.size == installations.length, "pkgx: env is being duped")
 
   const common_vars: Record<string, OrderedSet<string>> = {}
   const common_keys = new Set<string>()


mxcl:
I'm confused how this fixes things. If the assert was triggering before how were builds still continuing with a reduced env?

jhheider:
the assertions weren't triggering except in maybe 1 build (maybe? or zero.) but were preventing things like postgres and node from working in the same env.

however, the _fix_ basically broke `cmake` detection on line 69 and autoconf detection on line 73, since those both use `Set.has()`, and we changed the format of the set members.

mxcl:
cd failure https://github.com/pkgxdev/pkgx/actions/runs/10854509536/job/30125478243

jhheider:
both `./pkgx gpg-agent --daemon` and `bash -c './pkgx gpg-agent --daemon'` work in docker. so. frustrating.

jhheider:
even `echo 5 | cat | ./pkgx +sqlite3 gpg --import --batch --yes` works. so, this is another of the specie of "incomplete LD_LIBRARY_PATH", which isn't identified or fixed. i'll dig further.

jhheider:
```shell
pkgx/pkgx $ ./pkgx +gnupg.org | grep LD_LIB | tr : '\n'  | grep readline
/root/.pkgx/gnu.org/readline/v8.2.13/lib
```

mxcl:
yeah only fails for linux aarch64 and well, it's using the just built `pkgx` to execute `gpg` so suggests this fix isn't sufficient

mxcl:
The draft release could just be released without linux aarch64 since like, it's the least interesting platform

jhheider:
i can't prove it, but the only thing that changed for deps is the sync cache. so, db corruption, or an incomplete await, or something could have short data.

but deleting the deps in my pantry.db doesn't even affect my local invocation, so that's more mysterious. it's odd that `db.sql` is synchronous, but everything _looks like it should work_.

the fact that it's only messing up the LD_LIBRARY_PATH on linuxen is also concerning.

jhheider:
you know, the runners aren't clearing ~/.cache/pkgx; maybe I update their script and see.

jhheider:
that fixed it _immediately_. so it's cache corruption. fun. i bet `laravel` builds now.

jhheider:
couldn't be that easy, i guess: https://github.com/pkgxdev/pantry/actions/runs/10820951092/job/30130754367?pr=7239

mxcl:
If you can `ssh` in, what does `pkgx +php.net +gnu.org/gcc/libstdcxx` output?

mxcl:
Maybe downgrade `pkgx` in the actions for now.

jhheider:
> If you can `ssh` in, what does `pkgx +php.net +gnu.org/gcc/libstdcxx` output?

```shell
$ builder@f8b90684f45a:/$ pkgx +gnu.org/gcc/libstdcxx
 ✓  ~/.pkgx/gnu.org/mpc/v1.3.1
 ✓  ~/.pkgx/gnu.org/gmp/v6.3.0
 ✓  ~/.pkgx/gnu.org/mpfr/v4.2.1
 ✓  ~/.pkgx/gnu.org/gcc/libstdcxx/v14.2.0
 ✓  ~/.pkgx/gnu.org/binutils/v2.43.1
PATH="/home/builder/.pkgx/gnu.org/binutils/v2.43.1/bin${PATH:+:$PATH}"
CPATH="/home/builder/.pkgx/gnu.org/binutils/v2.43.1/include:/home/builder/.pkgx/gnu.org/gcc/libstdcxx/v14.2.0/include:/home/builder/.pkgx/gnu.org/mpc/v1.3.1/include:/home/builder/.pkgx/gnu.org/mpfr/v4.2.1/include:/home/builder/.pkgx/gnu.org/gmp/v6.3.0/include${CPATH:+:$CPATH}"
LIBRARY_PATH="/home/builder/.pkgx/gnu.org/binutils/v2.43.1/lib:/home/builder/.pkgx/gnu.org/gcc/libstdcxx/v14.2.0/lib:/home/builder/.pkgx/gnu.org/mpc/v1.3.1/lib:/home/builder/.pkgx/gnu.org/mpfr/v4.2.1/lib:/home/builder/.pkgx/gnu.org/gmp/v6.3.0/lib${LIBRARY_PATH:+:$LIBRARY_PATH}"
XDG_DATA_DIRS="/home/builder/.pkgx/gnu.org/binutils/v2.43.1/share:/home/builder/.pkgx/gnu.org/gcc/libstdcxx/v14.2.0/share:/home/builder/.pkgx/gnu.org/mpc/v1.3.1/share:/home/builder/.pkgx/gnu.org/mpfr/v4.2.1/share:/home/builder/.pkgx/gnu.org/gmp/v6.3.0/share${XDG_DATA_DIRS:+:$XDG_DATA_DIRS}"
MANPATH="/home/builder/.pkgx/gnu.org/binutils/v2.43.1/share/man:${MANPATH}:/usr/share/man"
PKG_CONFIG_PATH="/home/builder/.pkgx/gnu.org/mpfr/v4.2.1/lib/pkgconfig:/home/builder/.pkgx/gnu.org/gmp/v6.3.0/lib/pkgconfig${PKG_CONFIG_PATH:+:$PKG_CONFIG_PATH}"
LD_LIBRARY_PATH="/home/builder/.pkgx/gnu.org/binutils/v2.43.1/lib:/home/builder/.pkgx/gnu.org/gcc/libstdcxx/v14.2.0/lib:/home/builder/.pkgx/gnu.org/mpc/v1.3.1/lib:/home/builder/.pkgx/gnu.org/mpfr/v4.2.1/lib:/home/builder/.pkgx/gnu.org/gmp/v6.3.0/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
```

jhheider:
> Maybe downgrade `pkgx` in the actions for now.

probably an ok move

mxcl:
Seems to contain `/home/builder/.pkgx/gnu.org/gcc/libstdcxx/v14.2.0/lib` in the `LD_LIBRARY_PATH` setter.

jhheider:
yup, and even reverting to 1.1.6 doesn't solve it: https://github.com/pkgxdev/pantry/actions/runs/10891688927/job/30223144500

something strange up with x86-64. maybe the lib isn't in the right place.

jhheider:
nah, couldn't be that easy:
```shell
pkgx/work $ ls ~/.pkgx/gnu.org/gcc/libstdcxx/v14.2.0/lib/
gcc            libstdc++.so                libstdc++exp.la
libgcc_s.so    libstdc++.so.6              libstdc++fs.a
libgcc_s.so.1  libstdc++.so.6.0.33         libstdc++fs.la
libstdc++.a    libstdc++.so.6.0.33-gdb.py  libsupc++.a
libstdc++.la   libstdc++exp.a              libsupc++.la
```

jhheider:
```
/root/.pkgx/gnu.org/gcc/libstdcxx/v14.2.0/lib/libstdc++.so.6.0.33:        ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, with debug_info, not stripped
```

jhheider:
very mysterious.

PR #1052, opened 2024-11-25
fix(.nvmrc)

### LINKED ISSUE #1051: pkgx reads .nvmrc and can't parse lts/iron or ^20.0.0
```
$ dev
× unexpected error invalid semver range: @lts/iron
```

Here's the situation, we have `lts/iron` in `.nvmrc`. I tried adding `.node-version` and `pkgx:` to `package.json` but I still get the error. Then I tried changing `.nvmrc` to `^20.0.0` and I get the following error:

```
$ dev
× unexpected error invalid semver range: @^20.0.0
```

Next try gave:

```
$ dev
× unexpected error invalid semver range: @>20.0.0 <21.0.0
```

There's no mention of `.nvmrc` in the [dev docs](https://docs.pkgx.sh/using-dev/dev). Also, the error message that this is an invalid semver range is a little misleading (I think). It seems that it only accepts exact versions rather than ranges. Or maybe I misunderstood something.

It would be great if there was a clear precedence to pkgx files. I also tried adding `pkgx.yml` but that generates errors if both `.nvmrc` and `pkgx.yml` don't have the same version.

jhheider:
closes pkgxdev/dev#10


mxcl:
pls reopen at @pkgxdev/dev

jhheider:
replaced with https://github.com/pkgxdev/dev/pull/11

PR #1053, opened 2024-11-26
fix(install): fix unsafe install script (args passing)

rustdevbtw:
The unsafe install thing (that was added recently), creates a bit different script as you know.
Well, in that, the args that you pass to the program were being passed as a single argument (`"$ARGS"`). So, here's the simple one-line solution.

cc @jhheider

### DIFF
--- a/src/modes/install.ts
+++ b/src/modes/install.ts
@@ -71,7 +71,7 @@ export default async function(pkgs: PackageRequirement[], unsafe: boolean) {
                   # shellcheck source=$ENV_FILE
                   . "$ENV_FILE"
                   set +a
-                  exec "$PKGX_DIR/${pkgstr}/v*/bin/${program}" "$ARGS"
+                  exec "$PKGX_DIR/${pkgstr}/v*/bin/${program}" $ARGS
                 else
                   pkgx_resolve
                 fi


jhheider (commented):
Shell splitting is the worst. Will this work with spaces in arguments, or does the code really need to use an array?

rustdevbtw:
> Shell splitting is the worst. Will this work with spaces in arguments, or does the code really need to use an array?

Well, if you pass something like:
```sh
pandoc file.md -thtml
```
Using quotes will make it:
```
pandoc "file.kd -thtml"
```
So, it wouldn't work.

Without `$ARGS` quoted:
```
pandoc file.md -thtml
```

So, yeah. It'll work with spaces in arguments with this solution, I suppose.

rustdevbtw:
> Shell splitting is the worst. Will this work with spaces in arguments, or does the code really need to use an array?

Well, if you pass something like:
```sh
pandoc file.md -thtml
```
Using quotes will make it:
```
pandoc "file.md -thtml"
```
So, it wouldn't work.

Without `$ARGS` quoted:
```
pandoc file.md -thtml
```

So, yeah. It'll work with spaces in arguments with this solution, I suppose.

mxcl:
`pkgx` no longer installs, `pkgm` installs instead. Thank you and sorry!

If you want to install stubs we made a mash script where maybe this can go instead: https://github.com/pkgxdev/mash/blob/main/scripts/stub.sh

mxcl:
We could still merge this into the `v1` branch tho.

rustdevbtw:
> We could still merge this into the `v1` branch tho.

Then I'll create a PR for the v1 branch, since it's sorta a bug on that one.

rustdevbtw:
> > We could still merge this into the `v1` branch tho.
> 
> Then I'll create a PR for the v1 branch, since it's sorta a bug on that one.

By the way, I can't find the `v1` branch. Where is it?

jhheider:
i don't believe one exists. there is a `v1` tag at f28d9311eac07c0dd0b4c3900eab2420199b6604. we should have a branch there, in case there is demand for updates to v1.

PR #1070, opened 2025-01-14
Move installer out of Other Way to Install

abitrolly:
The installer is already referenced from the https://pkgx.sh and link to other methods would duplicate the info.


mxcl:
docs were rewritten and this no longer applies I think. LMK if wrong! Thanks for the contribution and apologies for the timing.

PR #1072, opened 2025-01-14
get coverage from running `pkgx`

mxcl:


mxcl:
Actually we're closer to 90%: https://coveralls.io/github/pkgxdev/pkgx

But for some reason we're getting penalized by a lack of coverage in our dependencies.

PR #1076, opened 2025-01-15
fix(install): fix unsafe install script (args passing)

rustdevbtw:
followup of #1053


mxcl:
https://github.com/pkgxdev/pkgx/releases/tag/v1.4.0

PR #1086, opened 2025-01-17
this fixes subprogs missing keys

### LINKED ISSUE #1085: getting `libcrypto.so.1.1 ... No such file or directory` for `pkgx npx ...`
Running pkgx 2.1 on kde plasma (zsh) and I keep getting `node: error while loading shared libraries: libcrypto.so.1.1: cannot open shared object file: No such file or directory` when trying to run `pkgx npx ..` or when trying to run nushell: `/home/eifr/.pkgx/nushell.sh/v0.101.0/bin/nu: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory`

jhheider:
closes #1085

### DIFF
--- a/crates/lib/src/env.rs
+++ b/crates/lib/src/env.rs
@@ -120,19 +120,12 @@ fn suffixes(key: &EnvKey) -> Option<Vec<&'static str>> {
 }
 
 pub fn mix(input: HashMap<String, Vec<String>>) -> HashMap<String, String> {
-    let mut rv = HashMap::new();
+    let mut rv = HashMap::from_iter(std::env::vars());
 
-    // append the existing keys
-    for (key, mut value) in std::env::vars() {
-        if let Some(injected_values) = input.get(&key) {
-            value = format!("{}:{}", injected_values.join(":"), value);
-        }
-        rv.insert(key, value);
-    }
-
-    // add the new keys
     for (key, value) in input.iter() {
-        if !rv.contains_key(key) {
+        if let Some(values) = rv.get(key) {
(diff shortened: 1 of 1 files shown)


mxcl:
Nice, I had a similar patch coming.

mxcl:
Again: unbelievable I didn’t spot this before! How on earth so many things I tested worked fine I have no idea.

jhheider:
i hate the two for-loops, but didn't feel like messing around with maps and zips and so forth.

jhheider:
my guess is because it's a script running under an execve that running in an interpreter... etc. levels on levels.

mxcl:
Ironically a couple hours ago I introduced a new test that exposes it: https://github.com/pkgxdev/pkgx/actions/runs/12835540088/job/35795376923?pr=1084#step:19:9

mxcl:
> i hate the two for-loops, but didn't feel like messing around with maps and zips and so forth.

Seems fine. We're iterating over two different things. Not sure there's a easy way to zip them together and do it clearly.

jhheider:
joyous. i guess the "right" way to do it is to clone `std::env::vars()`, then loop through `input` and insert_or_join each key in a single loop.

jhheider:
couldn't live with two loops. clone env, prepend needed keys.

mxcl:
Definitely need some unit tests for these pieces

jhheider:
that'd be an interesting one to spot regardless. needs a smart test. but preventing regressions should definitely be possible. an integration test stands a better chance.

PR #1090, opened 2025-01-24
Allow eg. `pkgx +rust` like pkgx^1 does

mxcl:


mxcl:
This leaves old `.db` files lying around but whatever.

I chose to rename so as to avoid a DB query every time pkgx runs just to determine if we need to upgrade. Maybe this is retarded. Maybe it's genius. Who knows?

crates/cli/src/main.rs
```diff
@@ -41,7 +41,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     let cache_dir = config.pantry_dir.parent().unwrap();
     std::fs::create_dir_all(cache_dir)?;
-    let mut conn = Connection::open(cache_dir.join("pantry.db"))?;
+    let mut conn = Connection::open(cache_dir.join("pantry.2.db"))?;
```
jhheider, line 44: i sense migrations coming. there's a fair few tools in rust for that. and a few in pkgx, like `goose`.

mxcl:
hmm it's just lucky (and probs wrong) that the display-name for rust is `rust` rather than `Rust`. Should make this case insensitive for the alias look up.

crates/lib/src/pantry_db.rs
```diff
@@ -103,6 +115,24 @@ pub fn which(cmd: &String, conn: &Connection) -> Result<Vec<String>, rusqlite::E
...
+        "
+        SELECT project FROM provides WHERE program = ?1
+        UNION
+        SELECT project FROM aliases WHERE alias = ?1;",
+    )?;
+    let mut rv = Vec::new();
+    let mut rows = stmt.query(params![symbol])?;
+    while let Some(row) = rows.next()? {
+        rv.push(row.get(0)?);
+    }
```
jhheider, line 132: just for style reasons, this is:
```rust
Ok(
    stmt
    .query(params![symbol])?
    .map(|row| row.get(0).expect("Query returned malformed data."))
    .collect::<Vec<String>
)
```
assuming the borrow checker doesn't think you're dropping bits.

jhheider (approved):

crates/lib/src/pantry_db.rs
mxcl: Doesn't compile and I am not enough of a rust guru yet to fix

crates/lib/src/pantry_db.rs
```diff
@@ -103,6 +115,24 @@ pub fn which(cmd: &String, conn: &Connection) -> Result<Vec<String>, rusqlite::E
...
 
+pub fn projects_for_symbol(
+    symbol: &String,
+    conn: &Connection,
+) -> Result<Vec<String>, rusqlite::Error> {
+    let mut stmt = conn.prepare(
+        "
+        SELECT project FROM provides WHERE program = ?1
+        UNION
+        SELECT project FROM aliases WHERE LOWER(alias) = LOWER(?1);"
```
jhheider, line 126: you could _store_ in `LOWER()` on the assumption that calling the function one time is cheaper. but you might discard data you want later.

crates/cli/src/main.rs
mxcl: we would certainly be more proper to be able to update the db incrementally and with migrations. That’s work I am not interested in RN tho for sure.

crates/lib/src/pantry_db.rs
jhheider: if you haven't, give `pkgx bacon -j clippy` a try ;)

crates/lib/src/pantry_db.rs
mxcl: yes this would be better

mxcl: thinking again, I guess we should store the correct representation from the yaml in case we need it that way some other time.

crates/lib/src/pantry_db.rs
mxcl: I have not, though I have been there numerous times lately since it is the latest addition to the pantry. I guess I don't see how it would help me here.

jhheider: not here specifically, but the feedback is rapid and, to my mind, better organized than rust-analyzer+vscode (it also works no matter what base directory you're in, which rust-analyzer struggles with).

PR #1092, opened 2025-01-25
Don’t absorb these errors

mxcl:


jhheider (approved):
Looks good. `Result<Option<T, E>>` feels some kind of way, but it makes sense.

PR #1104, opened 2025-01-29
Use XDG_DATA_HOME on *nix if no ~/.pkgx

### LINKED ISSUE #1103: XDG Compliance
It would be great if this package supported the `XDG` spec. Instead of cluttering the home directory with `~/.pkgx`, it can be tucked away under `~/.local/share/pkgx` or `${XDG_DATA_HOME:-$HOME/.local/share}/pkgx`. Please consider adding this feature, or allow us to specify the default PKGX location via environment variable.

mxcl:
Fixes #1103

### DIFF
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -42,18 +42,20 @@ fn get_pantry_dir() -> io::Result<PathBuf> {
 }
 
 fn get_pkgx_dir() -> io::Result<PathBuf> {
-    if let Ok(env_dir) = env::var("PKGX_DIR") {
-        let path = PathBuf::from(env_dir);
-        if !path.is_absolute() {
-            return Ok(env::current_dir()?.join(path));
-        } else {
+    if let Ok(path) = env::var("PKGX_DIR") {
+        let path = PathBuf::from(path);
+        if path.is_absolute() {
             return Ok(path);
         }
     }
-    #[cfg(target_os = "macos")]
-    return Ok(dirs_next::home_dir().unwrap().join(".pkgx"));
-    #[cfg(target_os = "linux")]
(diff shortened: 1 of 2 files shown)


crates/lib/src/config.rs
```diff
@@ -42,18 +43,28 @@ fn get_pantry_dir() -> io::Result<PathBuf> {
...
+    if let Ok(path) = env::var("PKGX_DIR").map(PathBuf::from) {
         if !path.is_absolute() {
-            return Ok(env::current_dir()?.join(path));
+            Ok(env::current_dir()?.join(path))
         } else {
-            return Ok(path);
+            Ok(path)
+        }
+    } else {
+        let home = dirs_next::home_dir();
```
jhheider, line 53: I'd love to just move the unwrapping here, since, if this fails, life is pretty much broken. But the `map()` calls below aren't wrong.

jhheider (approved):

crates/lib/src/config.rs
mxcl: I figure if for some reason `HOME` is unset but `XDG_BLAH` are we shouldn't fail here. We will fail later though.

mxcl:
```
Error: docs/faq.md:228:1 MD007/ul-indent Unordered list indentation [Expected: 0; Actual: 2]
```

This markdown linter is fucking junk. This is the third time it has reported completely fine markdown as invalid.

mxcl:
This is why I have no love for the markdown linting CI steps. I see almost no value, but it adds 2-4 minutes here and there because “reasons” that seem arbituary.

mxcl:
I’ll install the VSCode plugin

PR #1111, opened 2025-02-05
v1: uninstall if needed during install

felipecrs:
Closes https://github.com/pkgxdev/pkgm/issues/6

### DIFF
--- a/src/modes/install.ts
+++ b/src/modes/install.ts
@@ -1,6 +1,6 @@
 import { PackageRequirement, Path, PkgxError, hooks, utils } from "pkgx"
 import { is_shebang } from "../utils/get-shebang.ts"
-import { blurple, dim } from "../utils/color.ts"
+import uninstall from "./uninstall.ts"
 import undent from 'outdent'
 const { usePantry } = hooks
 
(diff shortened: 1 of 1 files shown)


felipecrs:
@mxcl I promise this is the last fix I need for pkgx v1. :)

felipecrs:
@mxcl, this is a gentle ping.

To test this PR, confirm it wasn't working before:

```
$ deno task run install openjdk.org@17 && deno task run install openjdk.org@21
# first installation succeeds
# second fails because openjdk.org@17 was already installed
```

Then, clone this PR and repeat.

mxcl:
I have disabled coveralls from failing PRs since they apparently can’t pick defaults that aren’t insane and despite many attempts to stop it being insane (by tweaking settings) its insanity prevails.

felipecrs:
> I have disabled coveralls from failing PRs since they apparently can’t pick defaults that aren’t insane and despite many attempts to stop it being insane (by tweaking settings) its insanity prevails.

I have the exact same problem using SonarQube for internal projects. It's just insanity, it never works well (comparing coverage of a PR against the coverage of target branch).

I'd personally avoid any tools like that and instead good old CLI tools that will scan the whole code base and return 0 or not. Some of them have deeper integration with GitHub Actions allow to provide highlight on the lines missing coverage directly in GitHub. I don't know anything about Rust tooling, though.

mxcl (approved):

felipecrs:
@mxcl this is a ping just in case you forgot to release a new patch version of pkgx v1. :)

felipecrs:
Right. Thank you very much. I'll consider everything when I switch, I haven't yet stopped to think about migrating to v2.

mxcl:
You do you. You never have to upgrade. pkgx^1 works fine and I don’t see that changing. A whole bunch of our infrastructure uses it and the only reason we may eventually switch is ^2 is a lot faster.

PR #1115, opened 2025-02-08
parent env can override pantry runtime env

### LINKED ISSUE #1101: eval "$(pkgx ...)" does not behave as expected
Will edit with more details, but basically: 

`set -a; eval "$(pkgx +blah)"; set +a; blah` does not always behave the same way as `pkgx +blah blah`

For example, `set -a; eval "$(pkgx +curl)"; set +a; curl https://google.com` produces a certificate error, while `pkgx +curl curl https://google.com` works fine (at least on my machine).

The problem seems to be due to the `SSL_CERT_FILE` environment variable, which becomes the same path twice in a row (with a separating colon). Something like: `/Users/user/.pkgx/stuff/cert.pem:/Users/user/.pkgx/stuff/cert.pem`. If you edit the variable to be just the single path `/Users/user/.pkgx/stuff/cert.pem` after `set +a`, curl begins to work on HTTPS links again.

The strange thing is though: if you run `pkgx +curl echo $SSL_CERT_FILE`, it does print out the double path. How does this work behind the scenes? Why does `eval "$(pkgx ...)"` behave differently from the one-liner?

mxcl:
If the runtime env does not recurse its key, parent env overrides.

Eg. SSL_CERT_FILE={{prefix}}, if the parent env sets SSL_CERT_FILE then it persists.

Most runtime env recurses, eg. PYTHONPATH={{prefix}}/lib/python:$PYTHONPATH, and in such cases the parent env is inserted (this is unchanged behavior).

Fixes #1101

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -12,6 +12,7 @@ use libpkgx::{
     config::Config, env, hydrate::hydrate, install_multi, pantry_db, resolve::resolve, sync,
     types::PackageReq, utils,
 };
+use regex::Regex;
 use rusqlite::Connection;
 use serde_json::json;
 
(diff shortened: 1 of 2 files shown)


mxcl:
```sh
$ pkgx +curl.se/ca-certs
SSL_CERT_FILE="${SSL_CERT_FILE:-/Users/mxcl/.pkgx/curl.se/ca-certs/v2024.12.31/ssl/cert.pem}"

$ pkgx +curl.se/ca-certs -- env
SSL_CERT_FILE=/Users/mxcl/.pkgx/curl.se/ca-certs/v2024.12.31/ssl/cert.pem

$ SSL_CERT_FILE=foo pkgx +curl.se/ca-certs -- env
SSL_CERT_FILE=foo
```

jhheider (approved):
Hairy, but it doesn't look wrong.

mxcl:
Yeah I hate using magic strings like this, but I am not fresh enough at 1am Friday to do better.

mxcl:
More interested in if you think the behavior is right or if I'm missing something.

jhheider:
I think the problem for certain keys is they're not arrayable by default. Most of the ones we set are, but I believe this one isn't. So, you have to decide which values to trust.

mxcl:
The pantry `runtime` key decides. If it recurses the key we inherit the env, if not the parent env overrides.

Issue #1100, opened 2025-01-26
`pkgx install python@3.11` - missing destination operand


abitrolly:
```sh
# curl https://pkgx.sh | sh
# pkgx install python@3.11
install: missing destination file operand after 'python@3.11'
Try '/root/.pkgx/gnu.org/coreutils/v9.6.0/bin/install --help' for more information.
```

Installed `pkgx` with `curl` into a fresh Fedora container.

```sh
# pkgx --version
pkgx 2.1.2
```

abitrolly:
@jhheider it just hangs indefinitely.

```
# curl -LIvvvv https://dist.pkgx.dev/pantry.tgz
* Host dist.pkgx.dev:443 was resolved.
* IPv6: (none)
* IPv4: 13.32.99.66, 13.32.99.69, 13.32.99.2, 13.32.99.46
*   Trying 13.32.99.66:443...
* Connected to dist.pkgx.dev (13.32.99.66) port 443
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/pki/tls/certs/ca-bundle.crt
*  CApath: none

```

jhheider:
very strange.

```shell
% curl -LIvvvv https://dist.pkgx.dev/pantry.tgz
* Host dist.pkgx.dev:443 was resolved.
* IPv6: (none)
* IPv4: 18.161.34.2, 18.161.34.78, 18.161.34.64, 18.161.34.29
*   Trying 18.161.34.2:443...
* Connected to dist.pkgx.dev (18.161.34.2) port 443
* ALPN: curl offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
* (304) (IN), TLS handshake, Server hello (2):
* (304) (IN), TLS handshake, Unknown (8):
* (304) (IN), TLS handshake, Certificate (11):
* (304) (IN), TLS handshake, CERT verify (15):
* (304) (IN), TLS handshake, Finished (20):
* (304) (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / AEAD-AES128-GCM-SHA256 / [blank] / UNDEF
* ALPN: server accepted h2
* Server certificate:
*  subject: CN=dist.tea.xyz
*  start date: Aug 27 00:00:00 2024 GMT
[...]

PR #1116, opened 2025-02-08
Install to tmpdir then atomic move

mxcl:
Refs #1100

### DIFF
--- a/crates/lib/src/install.rs
+++ b/crates/lib/src/install.rs
@@ -2,6 +2,7 @@ use async_compression::tokio::bufread::XzDecoder;
 use fs2::FileExt;
 use reqwest::Client;
 use std::{error::Error, fs::OpenOptions, path::PathBuf};
+use tempfile::tempdir_in;
 use tokio::task;
 use tokio_tar::Archive;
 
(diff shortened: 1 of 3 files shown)


crates/lib/src/install.rs
```diff
@@ -7,6 +7,7 @@ use tokio_tar::Archive;
 
 // Compatibility trait lets us call `compat()` on a futures::io::AsyncRead
 // to convert it into a tokio::io::AsyncRead.
+use tempfile::{self, tempdir_in};
```
jhheider, line 10: Probably shouldn't be between this comment and the following line.

crates/lib/src/install.rs
```diff
@@ -93,9 +94,16 @@ where
...
+    let temp_dir = tempdir_in(config.pkgx_dir.join(&pkg.project))?;
+
+    // Step 4: Extract the tar archive
     let mut archive = Archive::new(decoder);
-    archive.unpack(&config.pkgx_dir).await?;
+    archive.unpack(&temp_dir).await?;
+
+    // Step 5: atomically move from temp dir to installation location
+    let partial_path = format!("{}/v{}", pkg.project, pkg.version);
+    fs::rename(temp_dir.path().join(&partial_path), &dst_path)?;
```
jhheider, line 106: Quick google suggests this won't work across mounts. So, probably wants a comment to find it when someone triggers it.

mxcl: It'll be the same mount, we use `tempdir_in` to make the tmp file in `$PKGX_DIR`.

mxcl:
The failure seems to be due to the way the `tempfile` crate handles self-deleting the tempdir. https://docs.rs/tempfile/latest/tempfile/#early-drop-pitfall

But I am struggling to stop it doing it. It’s remarkable consistent when installing `git`. It’s always the openssl.org bottle that fails to get moved before the temp_dir is deleted.

mxcl:
lol it's because semverator renders the openSSL version as 1.1.1.23. My bad when implementing the render function probs.

PR #1125, opened 2025-02-13
Clone pantry to same location as pkgx^1

mxcl:
This change was thoughtless and accidental. It means we can end up with pkgx^1 and pkgx^2 getting out of sync.


mxcl:
This fix should be safe though the user will end up with a pantry clone that will never be used in the wrong place.

It also is more correct since the pantry clone is data and the db is cache.

PR #1128, opened 2025-02-14
Revert "Bump rusqlite, update rest within constraints"

mxcl:
This reverts commit b20b3f1511412062d3632e66709554c460dcf2f1.

The ca-cert embed worked, but then started failing when I bumped the deps. This was not caught in CI because the CI for the PR ran against `main` that didn't have the bump in it and the bump ran against main that didn't have the embed in it.

lol!

CI did [actually catch it](https://github.com/pkgxdev/pkgx/actions/runs/13319566842/job/37201556550) when it ran against the merge of the PR, but there is no mechanism to communicate that to anyone (nice).

Thus I released 2.3.0 and the embed doesn’t work.

`cd` does run “smoke tests” against `pkgx` before deploying. These tests are not run in a minimal container, so nothing failed.

Nice sequence for failure here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,4 +186,4 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - run: apt-get remove curl ca-certificates openssl --yes
       - run: cargo build
-      - run: .target/debug/pkgx +git
+      - run: ./target/debug/pkgx +git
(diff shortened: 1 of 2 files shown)


mxcl:
k passed after the revert. I will no revert the revert to verify failure.

mxcl:
lol ci won’t run because it is now the same as `main` I’m guessing.

jhheider:
you could just leave the environment's openssl in and set the OPENSSL_STATIC env-var (iirc), I believe. i _think_ that static links always, not just on self-build.

PR #1133, opened 2025-02-24
`pkgx --ensure make` uses the system make if poss

mxcl:


mxcl:
There are edge cases that this does not handle, ottomh: `pkgx -e +git` would not work.

jhheider:
> Also, importantly, `pkgx -e deno^2` will not check the version of the thing we are ensuring. So maybe this should not be added in fact.

this could range from `easy` to `fully impossible` to ensure, depending on the package in question. even parsing and hoping _if_ the package supports `--version|-V` would be fraught.

mxcl:
We could add something to package.yml to parse versions but that’s a lot of additional metadata for a feature that is only “nice”. So yeah your suggestion is better, though I am leaning towards this should be a mash script.

```sh
$ pkgx mash pkgx/fallback git clone […]
```

jhheider:
it's a lot of additional characters, but since it's maximally useful in scripts and README's, it's far from impossible.

mxcl:
Additionally, `foo --version` is hardly a well defined API though we could mandate a check when we build new versions to ensure we still can parse it ofc.

mxcl:
k I added `mash pkgx/ensure`, usage:

```sh
$ mash pkgx/ensure git --version
$ eval "$(mash pkgx/ensure +git)"
```

So I will revert this due to it not being possible to do properly with the full range of pkgx use. The mash script is much simpler too.

Issue #1136, opened 2025-03-02
An idea to speed up stubs


felipecrs:
Hi, I was doing some experiments, and I was able to speed up stubs by around 10%.

However, to make it work for real it will require a new flag or environment variable in `pkgx`, like `--swallow-first-arg` or `PKGX_SWALLOW_FIRST_ARG=1`.

The idea is simple: instead of relying on `sh` to run `pkgx`, use just `env` instead:

```console
❯ cat ./test-sh
#!/usr/bin/env sh

exec pkgx echo "$0" "$@"
❯ cat ./test-env
#!/usr/bin/env -S pkgx echo
```

Benchmarks:

```console
❯ pkgx hyperfine --warmup 2 --shell none --runs 1000 './test-sh' './test-env'
Benchmark 1: ./test-sh
  Time (mean ± σ):       7.1 ms ±   0.5 ms    [User: 3.9 ms, System: 4.1 ms]
  Range (min … max):     6.5 ms …  15.1 ms    1000 runs

Benchmark 2: ./test-env
  Time (mean ± σ):       6.5 ms ±   0.2 ms    [User: 3.9 ms, System: 3.5 ms]
  Range (min … max):     5.9 ms …   8.0 ms    1000 runs

Summary
  ./test-env ran
    1.09 ± 0.09 times faster than ./test-sh
```

But once the flag is implemented, then the stubs can look like:

```bash
#!/usr/bin/env -S pkgx --swallow-first-arg echo
```

Or something shorter like `pkgx -# echo`.

felipecrs:
Another option would be to implement some kind of pkgx interpreter, like:

```bash
#!/usr/bin/env pkgx

git@latest
```

pkgx would need to interpret this file as a list of args to pkgx itself and concatenate it user-provided args with it.

Maybe this would be less hacky, but maybe also slower, given pkgx would need to in fact open the file and read it.

jhheider:
that's cute. whether or not stubs have a lasting lifetime, using `pkgx` as a shebang is good. since this currently works:
```shell
#!/usr/bin/env -S pkgx +foo +bar
```
something like:
```shell
#!/usr/bin/env -S pkgx +foo +bar --exec foo
```

could work.

mxcl:
I like it, we could add this.

mxcl:
Why `-#`? Is there precedence? Note I am not dismissing it, just inquiring for research reasons.

felipecrs:
[...]

PR #1144, opened 2025-03-10
--shebang, -!

mxcl:
Refs #1136

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,6 +166,15 @@ jobs:
             -Xdemangler=rustfilt \
             > lcov.info
 
+      - name: --shebang test 1
+        run: |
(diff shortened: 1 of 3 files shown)


mxcl:
lol env interprets `--#!` so the # is a comment

jhheider:
fun fact, this works on both mac and linux:
`#!/usr/bin/env -S pkgx --shebang semverator`

but this _only_ works on mac:
`#!/usr/local/bin/pkgx --shebang semverator`

the latter gives this on linux (likely due to line splitting issues):
```shell
thread 'main' panicked at crates/cli/src/args.rs:81:22:
unknown argument --shebang semverator
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

felipecrs:
> the latter gives this on linux (likely due to line splitting issues):

This is crazy. That would mean yet a new flag to be added, like `env`'s `-S`.

PR #1146, opened 2025-03-11
Revamp docker image

felipecrs:
1. Update Debian from 10 to 12 (current latest)
2. Avoid installing apt-get dependencies, pkgx is meant to handle them all I suppose (maybe they were needed by pkgx v1?)
3. Remove bashrc craziness, they cause issues when running image as a different user like `--user $(id -u):$(id -g)` and I don't see why they are needed
4. Remove `CLICOLOR_FORCE=1`. This is non-sense. pkgx v2 already adds colors when the container is being run with `-it`.

The old image size was 176MB, the new image size is 86.6MB.

### DIFF
--- a/.github/.gitignore
+++ b/.github/.gitignore
@@ -1,0 +1,1 @@
+/products
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -1,16 +1,11 @@
-FROM debian:buster-slim AS stage0
+FROM debian:bookworm-slim AS base
+
+FROM base AS install
 COPY ./products/* /pkgx/
 RUN install -m 755 /pkgx/$(uname -m) /usr/local/bin/pkgx
 RUN install -m 755 /pkgx/pkgm /usr/local/bin/pkgm
-RUN echo 'export PS1="\\[\\033[38;5;63m\\]pkgx\\[\\033[0m\\]\\w $ "' >> /root/.bashrc
 
-FROM debian:buster-slim AS stage1
-RUN apt-get update && apt --yes install libc-dev libstdc++-8-dev libgcc-8-dev netbase libudev-dev ca-certificates
-COPY --from=stage0 /usr/local/bin/pkgx /usr/local/bin/pkgx
-COPY --from=stage0 /usr/local/bin/pkgm /usr/local/bin/pkgm
-COPY --from=stage0 /root/.bashrc /root/.bashrc
-ENV BASH_ENV=/root/.bashrc
-ENV CLICOLOR_FORCE=1
-SHELL ["/bin/bash", "-c"]
-CMD ["/bin/bash", "-i"]
+FROM base
+COPY --from=install /usr/local/bin/pkgx /usr/local/bin/pkgm /usr/local/bin/
 ENTRYPOINT ["/usr/local/bin/pkgx"]
+CMD ["/bin/bash"]


mxcl:
We'd have to update a lot of docs and a large amount of infra for these changes. What is your goal?

felipecrs:
> We'd have to update a lot of docs and a large amount of infra for these changes.

Can you point to some example, please? I thought I had not broken any use case.

felipecrs:
> For one the package we install are required packages on Linux. So dozens of pkgx packages will no longer work.

Can you point me to one? I tested running some crazy packages like Java and Python, and a few other, they all worked fine.

> What is your goal?

I want to use this image in some of my `docker-compose.yaml` to run maintenance scripts. Example:

```yaml
services:
  database:
    image: postgres
  maintainer:
    image: pkgxdev/pkgx:v2
    volumes:
      - ./scripts:/maintainer/scripts
    entrypoint:
      - bash
      - -c
      - set -eux; while true; do /maintainer/scripts/cleanup_db.sh; sleep 1h done;
```

And then in my script something like:

```bash
#!/bin/bash

set -o allexport
pkgx_env="$(pkgx +sometool@someversion +someothertool@someotherversion)"
eval "${pkgx_env}"
unset pkgx_env
set +o allexport

...
```

> What do you seek to do with these changes?

Well, to do what docker images are supposed to: allow for running applications (pkgx in this case) in a portable manner without having to install it.

Not just to run a simple command like for _demonstration_, but for more serious cases.

mxcl:
We will make a much leaner image when https://github.com/pkgxdev/manifests is ready. I am doing the work currently so that our packages only depend on glibc and nothing else.

felipecrs:
I see.

> But the fact is if you use pkgx and want to build eg. C++ these packages are required.

We could provide an additional variant of the pkgx docker image that is capable of building. Or maybe even make it as an image for brewkit instead of pkgx.

mxcl:
image

This is based on `busybox:1.31-glibc` which provides the glibc we build against.

glibc is forwards compatible so we could use the most recent glibc instead. busybox is tiny while still providing POSIX though there are caveats. In future would be fun to provide POSIX as `pkgx` shims. Not sure what image we should pick for now, but the above screenshot shows where we are at with a tiny 17MB image.

Here’s the `Dockerfile`:

```Dockerfile
FROM debian:buster-slim AS stage1
RUN apt-get update && apt-get install -y curl
RUN curl https://pkgx.sh | sh
RUN echo "#!/usr/bin/env -S pkgx -q! bash" > /bin/bash
RUN chmod +x /bin/bash

FROM busybox:1.31-glibc AS stage2
COPY --from=stage1 /usr/local/bin/pkgx /usr/local/bin/pkgx
COPY --from=stage1 /usr/local/bin/pkgm /usr/local/bin/pkgm
COPY --from=stage1 /lib/x86_64-linux-gnu/libgcc_s.so.1 /lib/libgcc_s.so.1
COPY --from=stage1 /lib/x86_64-linux-gnu/librt-2.28.so /lib/librt.so.1
COPY --from=stage1 /lib/x86_64-linux-gnu/libdl-2.28.so /lib/libdl.so.2
COPY --from=stage1 /usr/bin/env /usr/bin/env
COPY --from=stage1 /usr/bin/ldd /usr/bin/ldd
COPY --from=stage1 /bin/bash /bin/bash

SHELL ["/bin/sh", "-c"]
CMD ["/bin/sh", "-i"]
ENTRYPOINT ["/usr/local/bin/pkgx"]
```

felipecrs:
That's awesome. I personally don't like Busybox because it has no package manager to help adding more packages to the image, but maybe having pkgx is enough in that case.

mxcl:
We can provide multiple images easily enough. What base container would you like?

felipecrs:
I think having a Debian (Slim) or Ubuntu variant could be useful:

image

But maybe the best starter is the Alpine variant? It's marginally bigger than busybox.

I'm not asking for the Debian variant yet; I haven't found a specific need for it.

I think you should add more variants as needed. Maybe it will never be needed, who knows.

mxcl:
Alpine is no good until we have our own glibc package or we can find a good way to get glibc and libgcc into alpine. I looked into it and there was nothing trivial, but it's probably just a bit of work to do it

felipecrs:
Maybe go for busybox then. :)

Tag it with something like `pkgdev/pkgx:v2.5.0-busybox` if you don't want to commit future compatibility with the main variant yet.

mxcl:
https://github.com/pkgxdev/pkgx/pull/1157

felipecrs:
Superseded by #1157.

PR #1153, opened 2025-03-24
attempt to statically link libgcc on linux

mxcl:
Relevant conversation: https://github.com/rust-lang/rust/issues/29527

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -106,6 +106,18 @@ jobs:
 
       - run: |
           cargo build --release
+        if: ${{ runner.os != 'Linux' }}
+
+      # hack to statically link libgcc
+      - run: |
+          echo '#!/bin/bash' > linkerdriver.wrap
+          echo 'exec "${CC:-cc}" "${@/-lgcc_s/-lgcc_eh}"' >> linkerdriver.wrap
+          chmod +x ./linkerdriver.wrap
+
+          CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER="$PWD/linkerdriver.wrap" cargo build --release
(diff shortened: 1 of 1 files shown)


mxcl:
k well it worked:

```sh
$ ldd ./pkgx
	linux-vdso.so.1 (0x00007fff[4](https://github.com/pkgxdev/pkgx/pull/1153/checks#step:7:5)1294000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f77e1f60000)
	libpthread.so.0 => /lib/x8[6](https://github.com/pkgxdev/pkgx/pull/1153/checks#step:7:7)_64-linux-gnu/libpthread.so.0 (0x00007f77e1f3f000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x0000[7](https://github.com/pkgxdev/pkgx/pull/1153/checks#step:7:8)f77e087d000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f77e1f3a000)
	libc.so.6 => /lib/x[8](https://github.com/pkgxdev/pkgx/pull/1153/checks#step:7:9)6_64-linux-gnu/libc.so.6 (0x00007f77e06bd000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f77e1f6f000)
```

We now can only depend on glibc (all these libs are glibc + the kernel’s lib)

mxcl:
Note that *probably* we will just link musl and thus be dep free. However there are caveats to this I have found in my research, eg. https://users.rust-lang.org/t/static-linking-for-rust-without-glibc-scratch-image/112279/5

Where you need to patch the malloc implementation or you can lose performance massively. So basically it's not a trivial thing to try and I'm not ready yet.

PR #1157, opened 2025-03-26
more/new docker images

mxcl:


docker/Dockerfile.archlinux
```diff
@@ -0,0 +1,12 @@
...
+COPY ./products/* /pkgx/
+RUN install -m 755 /pkgx/$(uname -m) /usr/local/bin/pkgx
+RUN install -m 755 /pkgx/pkgm /usr/local/bin/pkgm
+
+FROM archlinux AS stage1
+COPY --from=stage0 /usr/local/bin/pkgx /usr/local/bin/pkgx
+COPY --from=stage0 /usr/local/bin/pkgm /usr/local/bin/pkgm
+
+SHELL ["/bin/bash", "-c"]
+CMD ["/bin/bash", "-i"]
```
felipecrs, line 11: `-i` should not be necessary as there's nothing useful in bashrc at this point. For example, `PATH` changes should be done at the image level.

`["/bin/bash"]` should be enough.

docker/Dockerfile.debian
```diff
@@ -0,0 +1,12 @@
+FROM debian AS stage0
+COPY ./products/* /pkgx/
+RUN install -m 755 /pkgx/$(uname -m) /usr/local/bin/pkgx
+RUN install -m 755 /pkgx/pkgm /usr/local/bin/pkgm
+
+FROM debian AS stage1
```
felipecrs, line 6: You should probably the slim debian variant.

docker/Dockerfile.ubuntu
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu AS stage0
+COPY ./products/* /pkgx/
+RUN install -m 755 /pkgx/$(uname -m) /usr/local/bin/pkgx
+RUN install -m 755 /pkgx/pkgm /usr/local/bin/pkgm
+
+FROM ubuntu AS stage1
```
felipecrs, line 6: I'd also recommend you pin the major version of the base image.

At least something like `ubuntu:noble` and `debian:bookwork-slim` to give you some control over it.

mxcl: I'm less keen to do this (at least for now, due to maintenance reasons). Users who want pinning should make their own images. Installing `pkgx` is trivial.

felipecrs: New Ubuntu and Debian major versions happen every 2 years... that shouldn't be a problem lol.

felipecrs: Yeah, but I meant how frequent the `ubuntu:latest` and `debian:latest` tag changes, which is Ubuntu LTS and Debian Stable.

mxcl: Really we should only offer images by glibc and kernel version, so I dunno why I’m even building ubuntu and debian images really.

jhheider: One comparison to think about it rust:

https://hub.docker.com/_/rust

their matrix seems to be

version = major|| major.minor || major.minor.patch
tag = version || version-os || os

mxcl: For now we are replacing the default image with `debian` which really is a breaking change, but we aren’t that big or important yet so I think we can afford to be naughty here.

We actually don’t use this image anywhere in our build infra—I was confused: we used to but no longer. We can fix anything I missed.

It will fail to work with c++ stuff however. Which is not good and means I won't merge yet. The new manifests I am working on currently do not depend on a pre-installed libstdc++, we build it. In general the new manifests are able to work with only glibc and libgcc. We can in future probs provide our own libgcc; I spent some time trying to get that working now and it was too painful. In further future we can also provide a glibc but that is also not a now thing.

felipecrs: > For now we are replacing the default image with `debian` which really is a breaking change, but we aren’t that big or important yet so I think we can afford to be naughty here.

I agree.

> It will fail to work with c++ stuff however.

I'd still like to reproduce myself. Do you mean running something like `pkgx gcc` would fail?

mxcl: Try building `node` with these images. It will fail.

felipecrs: I see now.

This could be a problem if some user uses or extends from the pkgx docker image to do things like building their own C++ projects.

Maybe we can consider that use case a stretch. Also, doing a `FROM pkgxdev/pkgx\nRUN apt-get udpate && apt-get install --no-install-recommends -y some-package` is probably a big deal for these users.

But I agree the best would be if it "just works", without having to install extra dependencies.

I'm glad to learn that this is being addressed in the new manifests, which would mean the containers need not to ship things like libstdc++. Very exciting. :)

mxcl: the previous image was debian yes, but it preinstalled some apt packages which the new one does not.

I’d like to be able to ship `pkgx` in alpine or even just a container with a kernel. But we need glibc and probably some logic in pkgx itself so it doesn't install glibc if it doesn't need to since that is excessive.

felipecrs: I completely agree. It would be awesome (and it would make a lot of sense) to have a distroless image with just pkgx.

PR #1162, opened 2025-04-01
Support `@latest`

### LINKED ISSUE #1160: I miss `@latest`
I noticed my `pkgx --version` was quite old so I updated it and it's broken most of my shell aliases which relied on `@latest` as a valid version annotation for always running the latest version of a package.

I saw there's `pkgx mash pkgx/upgrade` to update but this is a separate action that needs to be called before the thing I actually want to call which kinda defeats the point of me using pkgx at all instead of say, brew.

I understand if this separation of concern is the way you want this project to evolve but as of right now I'm sadly considering switching back to brew as pkgx doesn't really give me anything substantially better for my use case of "I want all my aliases to use the latest and greatest available versions of all my tools in one simple command".

mxcl:
Closes #1160


jhheider (approved):
Mmm, discriminated unions.

